PR #30, opened 2025-07-21
feat(frontend): create reusable loader component ss-19

D1monB:
This pull request adds a reusable Loader component that matches the overall app design
For testing purposes, the loader is temporarily added to the main page (App.tsx)


apps/frontend/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+type Properties = {
+	className?: string;
+	color?: string;
+	size?: number;
+};
+
+const AVERAGE_LOADER_SIZE = 32; // px
+
+const Loader: FC<Properties> = ({
```
liza-veis, line 13: let's use `({}: Properties) => JSX.Element` notation

apps/frontend/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,40 @@
+import { type FC } from "react";
+
+import styles from "./styles.module.css";
+
+type Properties = {
+	className?: string;
+	color?: string;
```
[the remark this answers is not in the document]
liza-veis: Yep, it would be nice to be able to have limited options, but I think for now, we only need one color, so we can remove this option

apps/frontend/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,40 @@
...
+};
+
+const AVERAGE_LOADER_SIZE = 32; // px
+
+const Loader: FC<Properties> = ({
+	className,
+	color = "black",
+	size = AVERAGE_LOADER_SIZE,
+}) => {
+	return (
```
liza-veis, line 18: Seems like this is a static picture, so let's use [vite-plugin-svgr](https://www.npmjs.com/package/vite-plugin-svgr) to load it as a component instead of storing its code in a component

apps/frontend/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,40 @@
+import { type FC } from "react";
+
+import styles from "./styles.module.css";
+
+type Properties = {
+	className?: string;
```
liza-veis, line 6: We will lean toward the approach of avoiding unnecessary customization. We will add specific fixed properties, e.g., size, when needed

```suggestion
```

liza-veis (changes requested):

liza-veis:
Let's use brand color (green like on buttons) instead of black, but overall yes

D1monB:
image

kalinkaaaa14 (approved):

liza-veis (approved):

liza-veis:
Let's wait for the #46 to be merged and reuse color value from there

Issue #18, opened 2025-07-21
feat: Generate authorization token


liza-veis:
### What feature?

Implement JWT token signing and verification.

https://www.npmjs.com/package/jose

- [ ] Add Token service to the backend with create and verify methods
- [ ] The token should store a user id
- [ ] The token should expire after 24h
- [ ] Return token in response to the sign-in and sign-up

liza-veis:
> 1. I assume I should use a symmetric secret implementation? If so, what secret string should I use? For now, I just took random key from CodeIgniter Encryption Keys section at [RandomKeygen](https://randomkeygen.com/). 

Yes, that's right, let's take it from env variables, e.g., JWT_SECRET. It will be our responsibility to generate appropriate value on the environment. For local development, you can put any string.

> 2. The `SignJWT` class has `setIssuer` and `setAudience` methods - should I use them? If yes, what values should I provide?

let's skip issuer and audience for now

> 3. `jwtVerify` can throw an error if token is invalid or expired. How should this error be handled - re-throw it from try/catch so some response middleware took care of it?

Let's keep the original error for the token service, we will catch this error and throw our custom error when we create authorization through this token

> 4. Another question about secret - as I understand, I need to save it in the .env file?

Yes

> Sorry if it's a lot of questions - it's my first time working with jose and second time working with JWT 😥

No worries about questions, it's good that you will have experience with something new for you 😎

mxm-kotyk:
@liza-veis 
I’ve finished working on this ticket and want to create a pull request. While I was working on it, tests for the auth controller and auth service were added. I updated those tests by adding token-related code, and now I’m not sure whether I should commit the changes from my branch (`18-feat-add-auth-token`), or if it would be better to commit them from the branch where the tests were originally added.

PR #42, opened 2025-07-21
feat(frontend): create header component ss-21

### LINKED ISSUE #21: feat: Create a header component
### What feature?

- [x] Create a header component
- [x] Default avatar should show two letters - the initials of the authenticated user (first name, last name)
- [x] Add header to the main page

### Screenshots

Image

vldkorin:
- Created `Header` component
- Implemented layout using `flexbox`
- used CSS variables for colors, font sizes, and font weights
Closes #21


apps/frontend/src/assets/css/variables.css
```diff
@@ -0,0 +1,43 @@
...
+	--color-text-disabled: #ecebeb;
+
+	/* Brand */
+	--color-brand-primary: #773800;
+	--color-brand-accent: #34a853;
+
+	/* Support */
+	--color-support-error: #ee4e2f;
+	--color-support-border: #bbb8b5;
+}
```
kalinkaaaa14, line 43: colors should be placed in the separate file

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,43 @@
...
+import styles from "./styles.module.css";
+
+type HeaderProperties = {
+	user: User;
+};
+
+type User = {
+	firstName?: string;
+	lastName?: string;
+};
```
kalinkaaaa14, line 12: authorized user will always have those fields included, make them required

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,43 @@
+import { type JSX } from "react";
+
+import styles from "./styles.module.css";
+
+type HeaderProperties = {
+	user: User;
+};
+
```
kalinkaaaa14, line 8:
```suggestion
type HeaderProperties = {
	user: User | null;
};

```

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,43 @@
...
+	const initials = getInitials(user);
+
+	return (
+		<header className={styles["header"]}>
+			<div className={styles["userInfo"]}>
+				<div className={styles["avatar"]}>{initials}</div>
+				<div className={styles["name"]}>
+					{user.firstName} {user.lastName}
+				</div>
+			</div>
```
kalinkaaaa14, line 38: You should create a separate Image component to display user avatar. This component should be reusable for other images

apps/frontend/src/assets/css/variables.css
```diff
@@ -0,0 +1,43 @@
...
+	--font-size-small: 16px;
+	--font-size-tiny: 14px;
+
+	/* Font line height */
+	--font-line-height-display: 64px;
+	--font-line-height-h1: 48px;
+	--font-line-height-h2: 40px;
+	--font-line-height-h3: 32px;
+	--font-line-height-small: 24px;
+	--font-line-height-tiny: 20px;
```
kalinkaaaa14, line 19: I don't think that line height should be placed as global variables. it will be ok to set line height directly where you need to without using variable

kalinkaaaa14 (approved):

kalinkaaaa14:
Also you should implement such logic for the header component

- if user is passed, the header should contain logo, avatar (initials or avatar url), first name + last name
- if user is not passed, then he should see the following header
image

kalinkaaaa14:
waiting for this issue to be merged
https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/issues/46

vldkorin:
It seems that I have updated everything according to the comments. 
I am waiting for the button component .
image
@kalinkaaaa14 @liza-veis

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,73 @@
...
+	const FIRST_INDEX = 0;
+
+	if (!user) {
+		return DEFAULT_INITIALS;
+	}
+
+	const firstInitial = user.firstName.charAt(FIRST_INDEX).toUpperCase();
+	const lastInitial = user.lastName.charAt(FIRST_INDEX).toUpperCase();
+
+	return firstInitial + lastInitial || DEFAULT_INITIALS;
```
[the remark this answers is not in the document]
liza-veis: done, please pull latest changes from main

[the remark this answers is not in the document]
vldkorin: > done, please pull latest changes from main

I guess I'm asking a lot of questions, I apologize for that. 
But I've already done something like this:
```
const getInitials = (user: User): string => {
    const [firstChar] = user.firstName;
    const [lastChar] = user.lastName;

    const firstNameInitial = firstChar ? firstChar.toUpperCase() : “”;
    const lastNameInitial = lastChar ? lastChar.toUpperCase() : “”;

    return `${firstNameInitial}${lastNameInitial}`;
};
```
If I do as Alina showed above, I get “object may be undefined” or something like that, so I had to do these checks
Should I change anything now?
@liza-veis

apps/frontend/src/libs/components/header/styles.module.css
```diff
@@ -0,0 +1,40 @@
...
+	justify-content: space-between;
+	width: 100%;
+	height: 72px;
+	padding: 0 32px;
+	border-bottom: 1px solid var(--color-stroke-weak);
+}
+
+.userInfo {
+	display: flex;
+	flex-shrink: 0;
```
liza-veis, line 14: hm, do we need this property 🤔 
```suggestion
```

apps/frontend/src/libs/components/header/styles.module.css
```diff
@@ -0,0 +1,40 @@
...
+	gap: 8px;
+	align-items: center;
+}
+
+.name {
+	font-family: var(--font-family);
+	font-size: var(--font-size-tiny);
+	font-weight: var(--font-weight-regular);
+	color: var(--color-text-strong);
```
liza-veis, line 23: let's avoid default values that are already present

```css
	font-family: var(--font-family); // let's set this in global css styles and only override for inputs/buttons
	font-size: var(--font-size-tiny); // same, it's just body text, so let's do this in global css
	font-weight: var(--font-weight-regular); // regular is default value
	color: var(--color-text-strong); // this one is also the default text color, so let's set it in global css
```

apps/frontend/src/libs/components/header/styles.module.css
```diff
@@ -0,0 +1,40 @@
...
+.logo {
+	display: flex;
+	gap: 8px;
+	align-items: center;
+}
+
+.logoIcon {
+	width: 136px;
+	height: 24px;
+}
```
liza-veis, line 35: It's better to define image width/height through tag attributes where possible, because this way it allocates needed space during render before styles load

apps/frontend/src/libs/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+	return `${firstNameInitial}${lastNameInitial}`;
+};
+
+const Avatar = ({ user }: Properties): JSX.Element => {
+	return (
+		<div className={styles["avatar"]}>
+			<Image
+				alt="User Avatar"
+				className={styles["avatarImage"] ?? ""}
```
liza-veis, line 31: let's use typecast to reconcile it with TypeScript. We will follow the next principle:

If this is the case, that can happen in code by its logic - we add js checks/safeguards. 

If by logic it's not possible or not expected, but typescript requires it - we will use typecast.

```suggestion
				className={styles["avatarImage"] as string}
```

apps/frontend/src/libs/components/avatar/styles.module.css
```diff
@@ -0,0 +1,29 @@
+.avatar {
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	width: 32px;
+	height: 32px;
```
liza-veis, line 6: Let's pass size as optional property through props with a default size of 32

apps/frontend/src/libs/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,41 @@
...
+	user: User;
+};
+
+type User = {
+	avatarUrl: null | string;
+	firstName: string;
+	lastName: string;
+};
+
+const getInitials = (user: User): string => {
```
liza-veis, line 16: Let's move this to a separate file following our usual structure - `avatar/libs/helpers/get-user-initials/get-user-initials.helper.ts`

vldkorin:
Please review the changes I have made.
image
@kalinkaaaa14 @liza-veis

apps/frontend/src/libs/components/header/styles.module.css
```diff
@@ -0,0 +1,28 @@
+.header {
+	box-sizing: border-box;
+	display: flex;
+	align-items: center;
+	justify-content: space-between;
+	width: 100%;
+	padding: 12px 32px 12px 48px;
```
liza-veis, line 7: Hm, why would header have different paddings on left/right? If you were trying to account for space in a user info, this padding should be applied to the user info container itself
```suggestion
	padding: 12px 32px;
```

apps/frontend/src/libs/components/header/styles.module.css
```diff
@@ -0,0 +1,28 @@
+.header {
+	box-sizing: border-box;
```
liza-veis, line 2: i believe in scaffolding styles we have this property applied to everything already

```suggestion
```

apps/frontend/src/libs/components/header/styles.module.css
```diff
@@ -0,0 +1,28 @@
...
+	display: flex;
+	align-items: center;
+	justify-content: space-between;
+	width: 100%;
+	padding: 12px 32px 12px 48px;
+	border-bottom: 1px solid var(--color-stroke-weak);
+}
+
+.userInfo {
```
liza-veis, line 11: Let's use kebab-case for CSS class names. The same is for avatar image or other class names 
```suggestion
.user-info {
```

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+type Properties = {
+	user: null | User;
+};
+
+type User = {
+	avatarUrl: null | string;
+	firstName: string;
+	lastName: string;
+};
```
liza-veis, line 17: I see you repeat this type at least a few times. Even though it's just a mock, let's move it to `header/libs/types/user.type.ts` and leave a todo comment to remove this interface when #18 is done. We will need to use AuthUserResponseDto here

apps/frontend/src/libs/components/avatar/styles.module.css
liza-veis: this comment is still not fixed

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,49 @@
...
+			</header>
+		);
+	}
+
+	return (
+		<header className={styles["header"]}>
+			<img alt="SmartScapes Logo" className={styles["logo"]} src={appLogo} />
+			<div className={styles["userInfo"]}>
+				<Avatar user={user} />
+				<div className={styles["name"]}>
```
liza-veis, line 41:
```suggestion
				<span className={styles["name"]}>
```

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,49 @@
...
+						<Button label="Sign in" type="button" />
+					</Link>
+				</div>
+			</header>
+		);
+	}
+
+	return (
+		<header className={styles["header"]}>
+			<img alt="SmartScapes Logo" className={styles["logo"]} src={appLogo} />
```
liza-veis, line 38: Add width and height attributes for logo, so it's controllable, and the browser will allocate specific space for it

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,49 @@
...
+					<Link to={AppRoute.SIGN_UP}>
+						<Button label="Sign up" type="button" />
+					</Link>
+					<Link to={AppRoute.SIGN_IN}>
+						<Button label="Sign in" type="button" />
+					</Link>
+				</div>
+			</header>
+		);
+	}
```
liza-veis, line 34: Since the structure is the same except for the right content, I believe you can just render it conditionally in one return statement, not to repeat the whole container. 
```jsx
const hasUser = Boolean(user);

return (
  // usual layout
 {hasUser ?
   <div /> :
  <div />
  }
  // rest of the layout
)
```

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,49 @@
...
+		return (
+			<header className={styles["header"]}>
+				<img alt="SmartScapes Logo" className={styles["logo"]} src={appLogo} />
+				<div className={styles["buttons"]}>
+					<Link to={AppRoute.SIGN_UP}>
+						<Button label="Sign up" type="button" />
+					</Link>
+					<Link to={AppRoute.SIGN_IN}>
+						<Button label="Sign in" type="button" />
+					</Link>
```
liza-veis, line 30: Hm, haven't we discussed updating the button component to have `to` property instead of wrapping it with link?

liza-veis (changes requested):

apps/frontend/src/libs/components/header/header.tsx
kalinkaaaa14: probably it was discussed here - https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/50

apps/frontend/src/libs/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,38 @@
...
+	firstName: string;
+	lastName: string;
+};
+
+const Avatar = ({
+	size = DEFAULT_AVATAR_SIZE,
+	user,
+}: Properties): JSX.Element => {
+	return (
+		<div className={styles["avatar"]} style={{ height: size, width: size }}>
```
kalinkaaaa14, line 25: we shouldn't use inline styles

vldkorin: > we shouldn't use inline styles

How can this be done differently so that when using this component, its size can be specified? Can I have a hint, please?

I thought about creating specific classes with sizes, but I don't think that would be convenient.
```
.avatar-40 {
	width: 40px;
	height: 40px;
}
```

Or this one but isn't it still inline style?
```
<div
	className={styles["avatar"]}
	style={{ "--avatar-size": `${size}px` } as React.CSSProperties}
>
```
```
.avatar {
	display: flex;
	align-items: center;
	justify-content: center;
	overflow: hidden;
	width: var(--avatar-size);
	height: var(--avatar-size);
	border-radius: 50%;
}
```

apps/frontend/src/libs/components/header/header.tsx
vldkorin: > Just merged this PR, so I think this part can be updated with the new to property in button component

It's already done

vldkorin: > Just merged this PR, so I think this part can be updated with the new to property in button component

Done

apps/frontend/src/libs/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,36 @@
...
+		<div
+			className={styles["avatar"]}
+			style={{ "--avatar-size": `${String(size)}px` } as React.CSSProperties}
+		>
+			<Image
+				alt="User Avatar"
+				className={styles["avatar-image"] as string}
+				fallback={
+					<span className={styles["fallback"]}>{getUserInitials(user)}</span>
+				}
```
h4vefundieyoung, line 29: probably image component should be refactored a little bit so we could get rid of these type casting. typescript gives you an opportunity to extend default img props interface - that's what I'd do.

apps/frontend/src/libs/components/header/styles.module.css
```diff
@@ -0,0 +1,33 @@
...
+	padding: 8px 16px;
+}
+
+.logo {
+	display: flex;
+	gap: 8px;
+	align-items: center;
+	max-width: 136px;
+	max-height: 24px;
+}
```
liza-veis, line 24: for fixed images like logo, prefer to set width and height through html attributes, not inline styles or css styles, so that browser can allocate space for them before styles are loaded. This only images which sizes are already defined 

```
<image width={136} height={24} />
```

And also since we now have logo as a separate image, do we need all those styles with display flex, gap etc? Seems like they are redundant

```suggestion
```

apps/frontend/src/libs/components/header/styles.module.css
```diff
@@ -0,0 +1,33 @@
...
+
+.buttons {
+	display: flex;
+	gap: 8px;
+}
+
+.buttons a {
+	width: auto;
+}
```
liza-veis, line 33: what are you trying to achieve with this 🤔 mostly we will avoid accessing elements through tags or anything except for class names, also we need to avoid side effects on common components (no nested access, except for libraries). If we need to modify specific properties of common components, we will do this explicitly through their interface as a component. e.g. passing isFullWidth or similar properties. But here it feels like there is no need for this, if you were trying to reduce buttons to their minimal size with this, it's better to just make sure buttons container takes the smallest width possible, e.g. buy aligning it to one of the corners through flex layout on a parent element

vldkorin: > what are you trying to achieve with this 🤔 mostly we will avoid accessing elements through tags or anything except for class names, also we need to avoid side effects on common components (no nested access, except for libraries). If we need to modify specific properties of common components, we will do this explicitly through their interface as a component. e.g. passing isFullWidth or similar properties. But here it feels like there is no need for this, if you were trying to reduce buttons to their minimal size with this, it's better to just make sure buttons container takes the smallest width possible, e.g. buy aligning it to one of the corners through flex layout on a parent element

if I remove width 100% from the button component, everything works fine. However, if I don't do this, the buttons stretch the container and have the wrong shape. The only thing that worked for me is what i had already done. So what should i do?
@liza-veis

liza-veis: What about adding `min-width: fit-content;` to the button component

apps/frontend/src/libs/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,39 @@
...
+	size?: number;
+	user: User;
+};
+
+const Avatar = ({
+	size = DEFAULT_AVATAR_SIZE,
+	user,
+}: Properties): JSX.Element => {
+	const avatarSource = user.avatarUrl?.trim() || undefined;
+	const hasAvatar = Boolean(avatarSource);
```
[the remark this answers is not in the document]
vldkorin: > then in getUserInitials(user) -> getUserInitials(firstName,lastName)
> 
> to avoid passing unnecessary data too much

I hope i did everything, left this
```
src={avatarUrl as string}
```
because I get
```
Type 'string | null' is not assignable to type 'string | undefined'.
  Type 'null' is not assignable to type 'string | undefined'
  ```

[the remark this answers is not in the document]
liza-veis: yep, we will be sticking to the following approach:

If, by the application logic, the case is possible, then we need to add a check or assign default values. If it's a TypeScript issue and the error is not possible, then we will cast the type

PR #43, opened 2025-07-21
feat(frontend): add error handling middleware with toast notifications ss-20

ZakZh:
- Implement error handling using redux middleware
- 
When an error occurs, show toast notification ([react-toastify](https://www.npmjs.com/package/react-toastify))
- Colors from the common colors in the design
- The error middleware handle only Redux errors (such as rejected async actions).
- Display whatever error message we receive in the response, but with a fallback.
- Toast notifications appear everywhere, including on auth/login pages.


apps/frontend/src/assets/css/variables.css
```diff
@@ -0,0 +1,2 @@
+/* color variables for the frontend application */
```
kalinkaaaa14, line 1: Redundant comment

apps/frontend/src/libs/modules/store/middlewares/error-handling.middleware.ts
```diff
@@ -0,0 +1,29 @@
...
+	(next: NextFunction) =>
+	(action: unknown): unknown => {
+		if (isRejected(action)) {
+			handleError(action as UnknownAction & { error: { message?: string } });
+		}
+
+		return next(action);
+	};
+
+const errorHandlingMiddleware: Middleware = () => handleAction;
```
kalinkaaaa14, line 27: using isRejected is not needed, it will be enough to check whether action.error exist and show a notification toaster, like this

```
const handleError: Middleware =
  () =>
  (next) =>
  (action): void => {
    if (action.error) {
      const { message } = action.error;
      notification.error(message);
    }

    return next(action);
  };
```

apps/frontend/src/libs/modules/store/middlewares/error-handling.middleware.ts
```diff
@@ -0,0 +1,29 @@
...
+} from "@reduxjs/toolkit";
+import { toast } from "react-toastify";
+
+type NextFunction = (action: unknown) => unknown;
+
+const handleError = (
+	action: UnknownAction & { error: { message?: string } },
+): void => {
+	const errorMessage = action.error.message || "An unexpected error occurred";
+	toast.error(errorMessage);
```
kalinkaaaa14, line 14: I think we will need some separate service on the frontend part that will have methods 
showError(message), showWarning(message), showSuccess(message), showInfo(message), that will wrap those calls from the react-tostify lib

liza-veis:
Please attach a screenshot of how the message looks.

apps/frontend/src/libs/components/toaster/styles.module.css
```diff
@@ -0,0 +1,39 @@
...
+:global(.Toastify__close-button) {
+	color: var(--color-text-strong);
+}
+
+:global(.Toastify__toast--error) {
+	color: var(--color-text-error);
+	background-image: linear-gradient(
+		var(--color-fill-error-weak),
+		var(--color-fill-error-weak)
+	);
```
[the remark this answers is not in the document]
ZakZh: Yeah, our common "--color-fill-..." variables have transparency, so you can see all the content underneath the toast. 
That’s why I decided to use a gradient instead.

ZakZh:
For now, since we don’t have the required colors, I left the default ones for Info and Warning
Знімок екрана 2025-07-23 225731

To match the example, I added an enum with a default error message, but I can remove it if needed
Знімок екрана 2025-07-23 231455

kalinkaaaa14 (approved):
lgtm

Issue #25, opened 2025-07-21
feat: Add a not found page


liza-veis:
### What feature?

- [x] Create not found page
- [x] User should be redirected to the not found page on an unknown route
- [x] Back to home button should lead user to the `/` route

Test result:

<img width="960" height="540" alt="Image" src="https://github.com/user-attachments/assets/22960c9d-d0fa-4f64-8471-6c57db075bb5" />

PR #44, opened 2025-07-22
feat(frontend): add common input component ss-28

h4vefundieyoung:
Common input component was implemented according to the design mockups.
Fonts styles and related dependencies will come from #25.

Input component can be tested through /sign-up route.

Screenshot 2025-07-22 at 04 34 48

https://github.com/user-attachments/assets/01f890c3-f343-4279-901c-83f6801b11ab


apps/frontend/src/libs/components/input/styles.module.css
```diff
@@ -1,14 +1,76 @@
...
+			var(--font-line-height-tiny) Inter,
+		sans-serif;
+	color: var(--text-error);
+}
+
+.label-caption {
+	font:
+		var(--font-weight-regular) var(--font-size-small) /
+			var(--font-line-height-small) Inter,
+		sans-serif;
```
kalinkaaaa14, line 24: better to split it to font-size, font-weight, font-family props

apps/frontend/src/assets/css/variables/colors.variables.css
```diff
@@ -0,0 +1,9 @@
+:root {
+	--stroke-strong: #004d3673;
+	--text-strong: #00291de5;
```
kalinkaaaa14, line 3: seems like the last version of colors is updated in slack conversion, so update it here also please

h4vefundieyoung: this vars story is pretty chaotic atm, probabaly i'll just pull it later from the main branch

apps/frontend/src/libs/components/input/styles.module.css
h4vefundieyoung: done

kalinkaaaa14:
LGTM, waiting for resolving this task
https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/issues/46

apps/frontend/src/assets/css/variables/colors.variables.css
h4vefundieyoung: merged

h4vefundieyoung:
> LGTM, waiting for resolving this task #46

main branch has been merged

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -33,15 +35,26 @@ const Input = <T extends FieldValues>({
...
+			{hasError && (
+				<span className={styles["error"]}>
+					<img
+						alt="error-icon"
+						className={styles["error-icon"]}
+						src={errorIcon}
+					/>
+					{error as string}
+				</span>
+			)}
```
liza-veis, line 57: We will need to replace this with a common icon component. Since it's most probably going to be added later in #45, we can leave it as it is for now. Just mentioning not to forget about this

apps/frontend/src/libs/components/input/styles.module.css
```diff
@@ -1,14 +1,75 @@
 .label {
 	display: flex;
 	flex-direction: column;
-	gap: 8px;
+	gap: 4px;
 	width: 100%;
-	color: var(--color-text-primary);
+	max-width: 364px;
```
liza-veis, line 6: We don't need max-width for inputs inside common input components. Inputs will be used inside forms or at least other containers, and it will be their responsibility to limit their width
```suggestion
```

apps/frontend/src/libs/components/input/styles.module.css
```diff
@@ -1,14 +1,75 @@
...
+	font-family: var(--font-family);
 }
 
 .error {
-	font-size: 14px;
-	font-weight: 500;
-	line-height: 1.3;
-	color: var(--color-danger);
+	display: inline-flex;
+	gap: 7px;
```
liza-veis, line 12: gaps, paddings, and usual sizes should be divisible by 4 from our design logic. If it's not this way on Figma, it was a mistake 😅 
```suggestion
	gap: 8px;
```

apps/frontend/src/libs/components/input/styles.module.css
```diff
@@ -1,14 +1,75 @@
...
+	align-items: center;
+	font-size: var(--font-size-tiny);
+	font-weight: var(--font-weight-semibold);
+	line-height: 20px;
```
liza-veis, line 16: I know in Figma line-height is set in px, but let's use unitless numbers (e.g., 1.5). Just divide the line-height px value by font size value and round it, e.g. `20/14=1.42 ~= 1.4`. And the same logic for other line-height mentions here

apps/frontend/src/libs/components/input/styles.module.css
```diff
@@ -1,14 +1,75 @@
...
+	font-size: var(--font-size-small);
+	font-weight: var(--font-weight-regular);
+	line-height: 24px;
+	color: var(--color-text-strong);
+}
+
+.label-caption::first-letter,
+.error::first-letter {
+	text-transform: capitalize;
+}
```
liza-veis, line 30: It's better to be controlled outside of the component
```suggestion
```

apps/frontend/src/libs/components/input/styles.module.css
```diff
@@ -1,14 +1,75 @@
...
+
+.error-icon {
+	width: 18px;
```
liza-veis, line 33: From the design, the size of the default icon is 24. If you exported it without the  frame around it, it's better to export icons with it, because this space normalizes them and makes icons scalable and uniform, especially when used in the same places as other icons

apps/frontend/src/libs/components/input/styles.module.css
```diff
@@ -1,14 +1,75 @@
...
+}
+
+.input {
+	height: 48px;
```
liza-veis, line 37: In our case, this height +- should be achieved by font-size + padding, no need to restrict it. Also, for text inside inputs, we want to see usual body-like text, so let's use font-size-small styles and respective line-height
```suggestion
```

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -15,7 +17,7 @@ type Properties<T extends FieldValues> = {
 	label: string;
 	name: FieldPath<T>;
 	placeholder?: string;
-	type?: "email" | "text";
+	type?: "email" | "password" | "text";
```
liza-veis, line 20: Let's pass option autoComplete property too, similar to type, for accessibility purposes

apps/frontend/src/libs/components/input/input.tsx
h4vefundieyoung: noted

timosh-yura:
LGTM

PR #45, opened 2025-07-22
feat(frontend): create sidebar component ss-22

nazartsya:
A sidebar component has been implemented. A sidebar has also been added to the main page and one item (Dashboard) has been added. The Dashboard leads to /.


apps/frontend/src/libs/components/sidebar/styles.module.css
```diff
@@ -0,0 +1,65 @@
...
+	text-decoration: none;
+}
+
+.sidebar-logo-icon {
+	width: 24px;
+	height: 24px;
+}
+
+.sidebar-logo-text {
+	font-size: 16px;
```
kalinkaaaa14, line 29: agreed for font size use rem, please apply the changes

liza-veis:
Please don't forget to attach issue to your PR

apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,39 @@
+import { Link } from "react-router";
+
+import dashboardIcon from "~/assets/images/dashboard.svg";
+import smartScapesLeafIcon from "~/assets/images/smart-scapes-leaf.svg";
+import { AppRoute } from "~/libs/enums/enums.js";
+
+import styles from "./styles.module.css";
+
+const Sidebar = (): React.JSX.Element => {
```
liza-veis, line 9: Sidebar should be a reusable component. We need to pass config from outside e.g. to pass items with the following interface

```typescript
type NavigationItem = {
	href: ValueOf<typeof AppRoute>;
	icon: IconName;
	label: string;
};
```

apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,39 @@
...
+				<img
+					alt="Smart Scapes Leaf"
+					className={styles["sidebar-logo-icon"]}
+					src={smartScapesLeafIcon}
+				/>
+				<span className={styles["sidebar-logo-text"]}>SmartScapes</span>
+			</div>
+			<ul className={styles["sidebar-navigation-list"]}>
+				<li>
+					<Link
```
liza-veis, line 22: Let's use NavLink  (should be exported from common components; we need to avoid library exports in multiple places). NavLink provides the ability to detect an active element which should be highlighted. Don't forget about accessibility and transitions. We need Sidebar to smoothly switch between options and have some hover effect on non-selected options

apps/frontend/src/assets/images/smart-scapes-leaf.svg
liza-veis, line 1: export logo fully with text, not just leaf

apps/frontend/src/assets/images/dashboard.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
liza-veis, line 1: let's store icons in dedicated folder - images/icons/dashboard.svg

liza-veis (changes requested):

apps/frontend/src/assets/images/smart-scapes-leaf.svg
nazartsya: Should the logo be a link to the main page?

nazartsya:
Can I use the clsx library?

apps/frontend/src/libs/components/sidebar/sidebar.tsx
liza-veis: yep, that's right

nazartsya:
demo

apps/frontend/src/libs/components/sidebar/styles.module.css
```diff
@@ -0,0 +1,27 @@
...
+	position: fixed;
+	top: 0;
+	left: 0;
+	z-index: 1;
+	display: flex;
+	flex-direction: column;
+	gap: 56px;
+	width: 360px;
+	height: 100%;
+	background-color: #ffffff;
```
h4vefundieyoung, line 11: can be shortened to #fff

timosh-yura: eslint doesnt allow shortening
image
and all colors in components should be assigned to variables

apps/frontend/src/libs/enums/icon-map.enum.ts
liza-veis, line 1: Maps and enums, even though similar, hold different logical values. Maps - map name with specific values, which could be anything, e.g., components in our case. Enums are a predefined enumeration of possible options. The only possible type of value of enum is string, which is a key of this specific value and corresponds to it, rather than maps to another value. So in our case, we need a map, which is right, but it should be stored in:

```
apps/frontend/src/libs/components/icon/libs/maps/icon-name-to-image.map.ts
```

And there will be something like: 
```
const iconNameToImage: Record<IconName, (props: React.SVGProps<SVGSVGElement>) => React.JSX.Element = {
	dashboard: Dashboard
};

export { iconNameToImage }
```

As for the IconName type, we definitely need it in common types, as it's already done right now, but let's keep it as a usual union type and use it in our map, instead of vice versa. 

```
type IconName = "dashboard" | "other-icon-name"
```

apps/frontend/src/libs/components/sidebar/styles.module.css
```diff
@@ -0,0 +1,27 @@
+.sidebar {
+	position: fixed;
+	top: 0;
+	left: 0;
+	z-index: 1;
+	display: flex;
+	flex-direction: column;
+	gap: 56px;
+	width: 360px;
```
liza-veis, line 9: It's 320 on the design and don't forget to avoid fixed values
```suggestion
	max-width: 320px;
```

apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,35 @@
...
+};
+
+const Sidebar = ({ navigationItems }: Properties): React.JSX.Element => {
+	return (
+		<div className={styles["sidebar"]}>
+			<NavLink
+				className={styles["sidebar-logo-link"] as string}
+				to={AppRoute.ROOT}
+			>
+				<Logo className={styles["sidebar-logo-icon"]} />
```
liza-veis, line 19: Let's use the logo as a usual image tag, and add height and width to it as HTML attributes, not through CSS. We need to prefer this style for static images, so that the browser can allocate needed space before layout is rendered

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -24,6 +26,19 @@ const App = (): React.JSX.Element => {
 
 	const isRoot = pathname === AppRoute.ROOT;
 
+	const navigationItems: NavigationItem[] = [
```
liza-veis, line 29: We may have a huge config with them and re-use in different navigation types, so let's move it to 

```
apps/frontend/src/libs/constants/navigation-items.constant.ts
```

and rename respectively
```
const NAVIGATION_ITEMS
```

apps/frontend/src/libs/components/link/link.tsx
```diff
@@ -17,3 +17,4 @@ const Link = ({ children, to }: Properties): React.JSX.Element => (
 );
 
 export { Link };
+export { NavLink } from "react-router";
```
liza-veis, line 20: let's export it from src/libs/components/components.ts. Ideally, each module exports a single value and related types, and all additional components, especially from libraries that we are planning to use in the app more than in one place, should be re-exported from src/libs/components/components.ts

apps/frontend/src/libs/components/nav-item/nav-item.tsx
```diff
@@ -0,0 +1,31 @@
+import { NavLink } from "~/libs/components/components.js";
+import { IconMap } from "~/libs/enums/enums.js";
+import { combineClassNames } from "~/libs/helpers/combine-class-names.helper.js";
+import { useCallback } from "~/libs/hooks/hooks.js";
+import { type NavigationItem } from "~/libs/types/types.js";
+
+import styles from "./styles.module.css";
+
+const NavItem = ({ href, icon, label }: NavigationItem): React.JSX.Element => {
```
liza-veis, line 9: let's call it SidebarItem and it should be a part of sidebar module

```
apps/frontend/src/libs/components/sidebar/libs/components/sidebar-item/sidebar-item.tsx
```

apps/frontend/src/libs/components/nav-item/nav-item.tsx
```diff
@@ -0,0 +1,31 @@
...
+	const Icon = IconMap[icon];
```
liza-veis, line 10: We need to create a common icon component that will accept height, name, and width, and we will map iconName to svg in this component

apps/frontend/src/libs/components/nav-item/styles.module.css
```diff
@@ -0,0 +1,34 @@
+.sidebar-navigation-link {
+	display: flex;
+	gap: 12px;
+	align-items: center;
+	padding: 12px 20px;
```
liza-veis, line 5: matching the design
```suggestion
	padding: 12px 24px;
```

apps/frontend/src/libs/components/nav-item/nav-item.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+	const getNavLinkClassName = useCallback(
+		({ isActive }: { isActive: boolean }): string =>
+			combineClassNames(
+				styles["sidebar-navigation-link"],
+				isActive && styles["sidebar-navigation-link-active"],
+			),
+		[],
+	);
```
liza-veis, line 19: let's simplify classnames a bit, e.g., icon, label. I remember I was mentioning it once in chat, that since we are using css modules and relatively small components, we need to give meaningful names to classes, but we don't need to clarify the context, because we already see it from the component name and uniqueness is guaranteed
```suggestion
	const getNavLinkClassName = useCallback(
		({ isActive }: { isActive: boolean }): string =>
			combineClassNames(
				styles["link"],
				isActive && styles["link-active"],
			),
		[],
	);
```

apps/frontend/src/libs/components/sidebar/styles.module.css
```diff
@@ -0,0 +1,27 @@
+.sidebar {
```
liza-veis, line 1: Hm, not sure about making it fixed, I would rather make it sticky if there are no obvious reasons to keep it outside of the layout, because it became harder to account for this space in common layout

apps/frontend/src/libs/components/sidebar/libs/components/sidebar-item/sidebar-item.tsx
```diff
@@ -0,0 +1,31 @@
+import { Icon, NavLink } from "~/libs/components/components.js";
```
liza-veis, line 1: This should be under the sidebar directory e.g. `components/sidebar/libs/components/sidebar-item`

apps/frontend/src/libs/components/icon/libs/maps/icon-name-to-svg.map.ts
```diff
@@ -0,0 +1,13 @@
+import Dashboard from "~/assets/images/icons/dashboard.svg?react";
+import ListOfPlaces from "~/assets/images/icons/list-of-places.svg?react";
```
liza-veis, line 2: Let's rename it to what this icon actually displays, because we may use it for other cases

liza-veis (changes requested):
Looks good overall, just a couple comments 😎

apps/frontend/src/libs/components/sidebar/libs/components/sidebar-item/sidebar-item.tsx
nazartsya: Do you mean create an index.ts file (apps\frontend\src\libs\components\sidebar\libs\components\sidebar-item\index.ts)?
And then in sidebar-item.tsx write import { Icon, NavLink, } from "~/libs/components/sidebar/libs/components/sidebar-item/index.js";?

apps/frontend/src/libs/components/icon/libs/maps/icon-name-to-svg.map.ts
nazartsya: maybe a second analytics icon)

apps/frontend/src/libs/components/components.ts
```diff
@@ -2,10 +2,13 @@ export { App } from "./app/app.js";
...
+export { Icon } from "./icon/icon.js";
 export { Input } from "./input/input.js";
 export { Link } from "./link/link.js";
 export { Loader } from "./loader/loader.js";
 export { RouterProvider } from "./router-provider/router-provider.js";
+export { Sidebar } from "./sidebar/sidebar.js";
 export { ToastContainer } from "./toast-container/toast-container.js";
 export { Provider as StoreProvider } from "react-redux";
 export { Outlet as RouterOutlet } from "react-router";
+export { NavLink } from "react-router";
```
kalinkaaaa14, line 14:
```suggestion
export { Outlet as RouterOutlet, NavLink } from "react-router";
```

apps/frontend/src/libs/components/sidebar/libs/components/sidebar-item/sidebar-item.tsx
kalinkaaaa14: Please, do not resolve comments left by other people, this is a responsibility of the reviewer 🙏 

Also, I think that that's not what Liza wanted, because we won't create .index files for such purposes, and in general, such files are mostly useless

@liza-veis please clarify what did you mean on this comment

apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,33 @@
...
+type Properties = {
+	navigationItems: NavigationItem[];
+};
+
+const Sidebar = ({ navigationItems }: Properties): React.JSX.Element => {
+	return (
+		<div className={styles["sidebar"]}>
+			<NavLink className={styles["logo-link"] as string} to={AppRoute.ROOT}>
+				<img alt="SmartScapes" height="24" src={smartScapesLogo} width="136" />
+			</NavLink>
```
kalinkaaaa14, line 18: We have a component Link which is exactly the wrapper on NavLink component. I think we should use Link component, and do not export NavLink at all

apps/frontend/src/libs/components/sidebar/libs/components/sidebar-item/sidebar-item.tsx
liza-veis: yes, agree. I mean to move this component to  `components/sidebar/libs/components/sidebar-item/sidebar-item.tsx`. This is our usual structure for sub-component (they should be nested inside their respective module folders)

liza-veis (approved):
Looks good 🐈

PR #54, opened 2025-07-22
feat(backend): generate authorization token ss-18

### LINKED ISSUE #18: feat: Generate authorization token

mxm-kotyk:
Resolves #18 

Right now the token is returned only in response to sign-up as there is no controller for sign-in.


apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -83,13 +83,16 @@ class AuthController extends BaseController {
 		options: APIHandlerOptions<{
 			body: UserSignUpRequestDto;
 		}>,
-	): Promise<APIHandlerResponse<UserSignUpResponseDto>> {
+	): Promise<
+		APIHandlerResponse<{ token: string; user: UserSignUpResponseDto }>
```
liza-veis, line 87: Let's create a type for this e.g. UserAuthReponseDto

packages/shared/src/modules/token/libs/types/user-auth-response-dto.type.ts
```diff
@@ -0,0 +1,6 @@
+type UserAuthResponseDto<UserData> = {
```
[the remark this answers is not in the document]
mxm-kotyk: Is that OK that `UserSignUpResponseDto` will lose `firstName` and `lastName` properties?

apps/backend/src/modules/token/token.service.ts
```diff
@@ -0,0 +1,47 @@
...
+		const { id } = payload as TokenPayload;
+
+		return id;
+	}
+
+	private generateSecret(): Uint8Array {
+		const secret = new TextEncoder().encode(this.config.ENV.AUTH.JWT_SECRET);
+
+		return secret;
+	}
```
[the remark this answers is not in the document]
mxm-kotyk: Done

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,20 +1,39 @@
...
+		this.tokenService = tokenService;
 		this.userService = userService;
 	}
 
 	public async signUp(
 		userRequestDto: UserSignUpRequestDto,
 	): Promise<UserSignUpResponseDto> {
-		return await this.userService.create(userRequestDto);
+		const { email, id } = await this.userService.create(userRequestDto);
+		const token = await this.generateToken(id);
```
liza-veis, line 30: How we generate tokens may differ from the methods where we use them, and the logic here is pretty straightforward, so it's better not to isolate it, just to pass the user ID. If we wanted to isolate it, it would be something more specific, e.g., generateUserAuthToken.
```suggestion
		const { email, id } = await this.userService.create(userRequestDto);
		const token = await this.tokenService.createToken({ userId: user.id });
```

liza-veis (commented):
Overall looks good already, just a few more comments 😎

packages/shared/src/modules/token/libs/types/user-auth-response-dto.type.ts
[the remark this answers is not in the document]
liza-veis: yep, looks good

apps/backend/src/libs/modules/token/base-token.module.ts
```diff
@@ -0,0 +1,39 @@
...
+
+		return jwt;
+	}
+
+	public async verify(jwt: string): Promise<TokenPayload["id"]> {
+		const { payload } = await jwtVerify(jwt, this.secret);
+		const { id } = payload as TokenPayload;
+
+		return id;
+	}
```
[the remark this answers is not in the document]
mxm-kotyk: Done

apps/backend/src/libs/modules/token/libs/types/token-payload.type.ts
```diff
@@ -0,0 +1,5 @@
+import { type JWTPayload } from "jose";
+
+type TokenPayload = JWTPayload & { payload: unknown };
+
+export { type TokenPayload };
```
liza-veis, line 5: let's just do it this way:

`export { JWTPayload as TokenPayload } from 'jose'`

And instead of JWTPayload from the library, we will use TokenPayload. Also, do we really need to override this type and add a payload to it? Can't we just use JWTPayload as the payload itself

apps/backend/src/libs/modules/token/base-token.module.ts
```diff
@@ -0,0 +1,47 @@
...
+
+		return jwt;
+	}
+
+	public async verify<Payload extends JWTPayload = JWTPayload>(
+		jwt: string,
+	): Promise<Payload> {
+		const { payload } = await jwtVerify(jwt, this.secret);
+
+		return payload as Payload;
```
liza-veis, line 43: I'm curious if there is a posibility to pass it like this instead of casting, if no, then we will leave it this way
```suggestion
		const { payload } = await jwtVerify<Payload>(jwt, this.secret);

		return payload;
```

liza-veis (changes requested):

apps/backend/src/libs/modules/token/base-token.module.ts
mxm-kotyk: You are right - it works.
Done

lavandosya (approved):

PR #55, opened 2025-07-22
feat(backend): add groups and permissions tables and seeds ss-49

D1monB:
- This pull request adds new tables for groups and permissions in the database.
- It also adds a new column group_id to the users table.
- A seed file was created to add starting data: groups (admins, users) and one permission (create_route) for admins.

image
image
image
image


apps/backend/src/libs/modules/database/libs/enums/database-table-name.enum.ts
```diff
@@ -1,5 +1,8 @@
 const DatabaseTableName = {
+	GROUP_TO_PERMISSIONS: "group_to_permissions",
```
kalinkaaaa14, line 2:
```suggestion
	GROUPS_TO_PERMISSIONS: "groups_to_permissions",
```

apps/backend/src/db/seeds/01-seed-groups-and-permissions.ts
kalinkaaaa14, line 1: adding the permanent data should be done through another migration, so this config for seeds is redundant. 
seeding mostly is needed for temporal data, like for tests, etc, and that's not our case

so create another migration insert_groups_and_permissions_data where on up - the data will be inserted, on down - the same data will be deleted (find by the keys)

apps/backend/src/libs/modules/database/libs/enums/database-table-name.enum.ts
D1monB: Can I leave it? I added "src/db/seeds/*.ts" to knip.config.js so that the linter doesn't flag it as unused.

D1monB:
image
image
image
image

apps/backend/src/db/migrations/20250723031408_add_groups_and_permissions_tables.ts
```diff
@@ -0,0 +1,91 @@
+import { type Knex } from "knex";
+
+import { DatabaseTableName } from "~/libs/modules/database/libs/enums/enums.js";
```
kalinkaaaa14, line 3: As we've discussed on the daily today, importing some enums/constants etc. into the migration is a bad practice because migrations should be isolated from the application. This isolation ensures that migrations stay stable and reproducible over time, even if the application codebase evolves. So please create an enum with table names here directly in the migration file, similar as you did for column names

D1monB: So should I remove GROUPS, PERMISSIONS and GROUPS_TO_PERMISSIONS from enum ?
image

apps/backend/src/db/migrations/20250723031408_add_groups_and_permissions_tables.ts
```diff
@@ -0,0 +1,93 @@
+import { type Knex } from "knex";
+
+const TableNames = {
```
liza-veis, line 3: enums should be named in singular form
```suggestion
const TableName = {
```

apps/backend/src/db/migrations/20250723031408_add_groups_and_permissions_tables.ts
```diff
@@ -0,0 +1,93 @@
...
+} as const;
+
+async function down(knex: Knex): Promise<void> {
+	await knex.schema.alterTable(TableNames.USERS, (table) => {
+		table.dropColumn(ColumnName.GROUP_ID);
+	});
+
+	await knex.schema.dropTableIfExists(TableNames.GROUPS_TO_PERMISSIONS);
+	await knex.schema.dropTableIfExists(TableNames.PERMISSIONS);
+	await knex.schema.dropTableIfExists(TableNames.GROUPS);
```
liza-veis, line 27: Multiple knex calls should be wrapped in a transaction, the same for the up method and another migration.

apps/backend/src/db/migrations/20250723130000_insert_groups_and_permissions_data.ts
```diff
@@ -0,0 +1,95 @@
...
+		.where(ColumnName.KEY, "create_route")
+		.delete();
+
+	await knex(TableNames.GROUPS)
+		.whereIn(ColumnName.KEY, ["admins", "users"])
+		.delete();
+}
+
+async function up(knex: Knex): Promise<void> {
+	const now = new Date();
```
liza-veis, line 53: use knex.fn.now(), execution environment and database dates may differ

apps/backend/src/db/migrations/20250723130000_insert_groups_and_permissions_data.ts
```diff
@@ -0,0 +1,95 @@
...
+			[ColumnName.NAME]: "Admins",
+			[ColumnName.UPDATED_AT]: now,
+		},
+		{
+			[ColumnName.CREATED_AT]: now,
+			[ColumnName.KEY]: "users",
+			[ColumnName.NAME]: "Users",
+			[ColumnName.UPDATED_AT]: now,
+		},
+	]);
```
liza-veis, line 68: This way, you don't need to query them again 
```suggestion
const insertedGroups =	await knex(TableNames.GROUPS).insert([
		{
			[ColumnName.CREATED_AT]: now,
			[ColumnName.KEY]: "admins",
			[ColumnName.NAME]: "Admins",
			[ColumnName.UPDATED_AT]: now,
		},
		{
			[ColumnName.CREATED_AT]: now,
			[ColumnName.KEY]: "users",
			[ColumnName.NAME]: "Users",
			[ColumnName.UPDATED_AT]: now,
		},
	]).returning(["id", "key"]);
```

apps/backend/src/db/migrations/20250723130000_insert_groups_and_permissions_data.ts
```diff
@@ -0,0 +1,95 @@
...
+
+	const [group] = (await knex(TableNames.GROUPS)
+		.select(ColumnName.ID)
+		.where(ColumnName.KEY, "admins")) as IdObjectList;
+
+	const [permission] = (await knex(TableNames.PERMISSIONS)
+		.select(ColumnName.ID)
+		.where(ColumnName.KEY, "create_route")) as IdObjectList;
+
+	if (group && permission) {
```
liza-veis, line 85: If the values are missing, something definitely went wrong, and we would prefer migration to fail with an error and rollback changes. If you wrap it with a transaction, it will be done automatically
```suggestion
```

liza-veis (changes requested):

kalinkaaaa14 (approved):

liza-veis:
> I added const group = insertedGroups.find((g) => g.key === "admins"); Should I remove it? I don’t understand if the insertedGroups array always has the item with key 'admins' in the first position Because i could write const [group] = (await trx(TableName.GROUPS) .insert([ { [ColumnName.CREATED_AT]: trx.fn.now(), [ColumnName.KEY]: "admins", [ColumnName.NAME]: "Admins", [ColumnName.UPDATED_AT]: trx.fn.now(), }, { [ColumnName.CREATED_AT]: trx.fn.now(), [ColumnName.KEY]: "users", [ColumnName.NAME]: "Users", [ColumnName.UPDATED_AT]: trx.fn.now(), }, ]) .returning(["id", "key"])) without additional function

missed that comment 😅 I believe it has the same order and can be simplified, but it's better to check out in the knex docs. Anyway, as it's been merged, we will stay with this implementation.

PR #59, opened 2025-07-23
feat(backend/shared): add POIs module ss-52

### LINKED ISSUE #52: feat: Module to work with POIs
Create a separate module on the be to work with POI(points of interests) entity

- [x] Add new table to DB 

```
Table points_of_interest {
  id integer [primary key]
  created_at timestamp
  updated_at timestamp
  name varchar
  longitude float
  latitude float
}
```

- [x] Add `/points-of-interests` controller
- [x] Implement CRUD operations
- [x] Cover new functionality with tests

vyacheslavKhoroshev:
- Added full Points of Interest (POI) module with complete CRUD functionality;

- Includes custom error handling (PointOfInterestError) with clear messages;

- Service layer: checks for unique POI names, validates existence of POI by ID;

- Entity: encapsulates POI data using factory methods, encapsulates POI data using factory methods;

- Repository: handles database operations via Objection.js;

- Service: contains business logic and validation,;

- Controller: defines HTTP endpoints with proper status codes, added Swagger data;

- Added unit tests for repository and services;

Resolves #52


apps/backend/src/modules/points-of-interest/libs/exceptions/point-of-interest.exception.ts
h4vefundieyoung, line 1: yagni?

liza-veis: We will be using custom errors per service to improve error tracking (you will see it as a type of error you receive in the error stack), + we may modify them in the future with custom names. So let's leave it for now

apps/backend/src/modules/points-of-interest/points-of-interest.controller.ts
```diff
@@ -0,0 +1,274 @@
...
+
+		this.addRoute({
+			handler: this.find.bind(this),
+			method: HTTPMethodEnum.GET,
+			path: "/:id",
+		});
+
+		this.addRoute({
+			handler: this.findAll.bind(this),
+			method: HTTPMethodEnum.GET,
```
liza-veis, line 68: Since we are using HTTPMethod as a union type, I believe addRoute is already a safeguard from passing wrong values, and you can just use 'GET'. It's useful to have safeguards on everything in pure js, but in ts we will be using union types, unless we need specific values for computation, template string, listing values, etc.

apps/backend/src/modules/points-of-interest/points-of-interest.service.ts
```diff
@@ -0,0 +1,135 @@
...
+				name,
+			}),
+		);
+
+		const object = item.toObject();
+
+		return {
+			...object,
+			id: object.id,
+		};
```
liza-veis, line 38: It seems like toObject already returns value with id and the same for methods below, or is there a purpose for this?
```suggestion
		return item.toObject()
```

apps/backend/src/modules/points-of-interest/points-of-interest.service.ts
```diff
@@ -0,0 +1,135 @@
...
+		const existingPointOfInterest =
+			await this.pointsOfInterestRepository.findByName(name);
+
+		if (existingPointOfInterest) {
+			throw new PointOfInterestError({
+				message: PointOfInterestExceptionMessage.NAME_ALREADY_EXISTS,
+				status: HTTPCode.CONFLICT,
+			});
+		}
+	}
```
liza-veis, line 132: I like the approach, and it could be a possible implementation, but there is one thing to keep in mind, that mostly we want to do as less queries to the database as possible because they are very time and resource consuming.  In case of ensureIdExists, we can easily detect this without an additional query in most cases, e.g., in get - we just we will just get null, in update - we can retrieve a number of rows that were updated, in delete - we can also retrieve rows affected. So it will be better to handle those cases separately and just throw the same error. As for ensureNameIsUnique, I think we can leave it.

packages/shared/src/modules/points-of-interest/libs/validation-schemas/point-of-interest-update.validation-schema.ts
```diff
@@ -0,0 +1,31 @@
...
+				message: PointsOfInterestValidationMessage.NAME_MAXIMUM_LENGTH,
+			})
+			.optional(),
+	})
+	.refine(
+		(data) => {
+			return Object.keys(data).length >= MINIMUM_FIELDS_REQUIRED;
+		},
+		{
+			message: "At least one field must be provided for update",
```
liza-veis, line 27: use validation message enum for this, and validation rule for MINIMUM_FIELDS_REQUIRED, something like MINIMUM_FIELDS_TO_UPDATE

apps/backend/src/modules/points-of-interest/points-of-interest.repository.ts
```diff
@@ -0,0 +1,88 @@
...
+		return PointsOfInterestEntity.initialize(pointOfInterest);
+	}
+
+	public async delete(id: number): Promise<boolean> {
+		const deletedCount = await this.pointsOfInterestModel
+			.query()
+			.deleteById(id)
+			.execute();
+
+		return !!deletedCount;
```
liza-veis, line 34: let's prefer a more explicit cast instead of !!, just a common pattern used in the app
```suggestion
		return Boolean(deletedCount);
```

apps/backend/src/modules/points-of-interest/points-of-interest.controller.ts
```diff
@@ -0,0 +1,274 @@
...
+			path: "/",
+		});
+
+		this.addRoute({
+			handler: this.update.bind(this),
```
liza-veis, line 73: let's call it patch and the same for pointOfInterestPatchValidationSchema, not to be confused later
```suggestion
			handler: this.patch.bind(this),
```

liza-veis (changes requested):
I'm leaving a few comments, but good job overall, you seem to grasp conventions quickly 😎

apps/backend/src/modules/points-of-interest/points-of-interest.repository.ts
```diff
@@ -0,0 +1,94 @@
...
+				name,
+			})
+			.where("id", "=", id)
+			.returning("*")
+			.execute();
+
+		if (updatedPoints.length === 0 || !updatedPoints[0]) {
+			return null;
+		}
+
```
liza-veis, line 89: i believe it could be simplified to this
```suggestion
		const [updatedRow] = await this.pointsOfInterestModel
			.query()
			.patch({
				name,
			})
			.where("id", "=", id)
			.returning("*")
			.execute();
		
		

		if (!updatedRow) {
			return null;
		}

```

apps/backend/src/modules/points-of-interest/points-of-interest.service.ts
```diff
@@ -0,0 +1,112 @@
...
+			PointsOfInterestEntity.initializeNew({
+				name,
+			}),
+		);
+
+		return item.toObject();
+	}
+
+	public async delete(id: number): Promise<boolean> {
+		const deleted = await this.pointsOfInterestRepository.delete(id);
```
liza-veis, line 37: boolean properties should be named using is/has/should prefixes
```suggestion
		const isDeleted = await this.pointsOfInterestRepository.delete(id);
```

packages/shared/src/modules/points-of-interest/libs/validation-schemas/point-of-interest-update.validation-schema.ts
liza-veis: this comment is still not resolved

liza-veis (changes requested):
A few more comments, otherwise than that LGTM

apps/backend/src/modules/points-of-interest/points-of-interest.repository.ts
```diff
@@ -0,0 +1,86 @@
...
+	public async delete(id: number): Promise<boolean> {
+		const deletedCount = await this.pointsOfInterestModel
+			.query()
+			.deleteById(id)
+			.execute();
+
+		return Boolean(deletedCount);
+	}
+
+	public async find(id: number): Promise<null | PointsOfInterestEntity> {
```
lavandosya, line 37:
```suggestion
	public async findById(id: number): Promise<null | PointsOfInterestEntity> {
```

apps/backend/src/modules/points-of-interest/points-of-interest.service.ts
```diff
@@ -0,0 +1,112 @@
...
+			throw new PointOfInterestError({
+				message: PointOfInterestExceptionMessage.ID_NOT_FOUND,
+				status: HTTPCode.NOT_FOUND,
+			});
+		}
+
+		return true;
+	}
+
+	public async find(id: number): Promise<PointsOfInterestResponseDto> {
```
lavandosya, line 49:
```suggestion
	public async findById(id: number): Promise<PointsOfInterestResponseDto> {
```

packages/shared/src/libs/modules/http/libs/types/http-method.type.ts
```diff
@@ -1,3 +1,3 @@
-type HTTPMethod = "GET" | "POST";
+type HTTPMethod = "DELETE" | "GET" | "PATCH" | "POST";
```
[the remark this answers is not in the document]
vyacheslavKhoroshev: Can't fix this because of lint sorting error

packages/shared/src/modules/points-of-interest/libs/validation-schemas/point-of-interest-create.validation-schema.ts
```diff
@@ -0,0 +1,21 @@
+import { z, type ZodType } from "zod";
+
+import {
+	PointsOfInterestValidationMessage,
+	PointsOfInterestValidationRule,
+} from "../enums/enums.js";
+import { type PointsOfInterestRequestDto } from "../types/types.js";
+
+const pointOfInterestCreate: ZodType<PointsOfInterestRequestDto> =
```
kalinkaaaa14, line 9: the type should be inferred from zod library, and there is no need to type it here
```suggestion
const pointOfInterestCreate
```

liza-veis (approved):
Good work 🙌

PR #67, opened 2025-07-24
feat(backend/shared): add sign in flow to the backend ss-17

my-derkach:
Implemented sign-in flow to the backend:
- add /auth/sign-in endpoint to the auth controller
- add sing in method to the auth service
- return user as response without password
- add sign in request dto
- add sign in validation


apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,14 +1,50 @@
+import { ApplicationError } from "@smartscapes/shared/src/libs/exceptions/exceptions.js";
+
+import { type encryption } from "~/libs/modules/encryption/libs/encription.js";
```
h4vefundieyoung, line 3: i know it's not yours, but i'd suggest to rename it to encryption. there are a few typo files like this.

liza-veis: agree let's rename it, but please use `git mv` for this, so that git treats this file as moved, not created

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,14 +1,50 @@
...
+
+		const user = await this.userRepository.findByEmail(email);
+
+		if (!user) {
+			throw new ApplicationError({ message: "Invalid credentials." });
+		}
+
+		const passwordsMatch = await this.encryptionService.compare(
+			password,
+			user.getPasswrodHash(),
```
lavandosya, line 40:
```suggestion
			user.getPasswordHash(),
```

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,14 +1,50 @@
...
+	): Promise<UserSignUpResponseDto> {
+		const { email, password } = userRequestDto;
+
+		const user = await this.userRepository.findByEmail(email);
+
+		if (!user) {
+			throw new ApplicationError({ message: "Invalid credentials." });
+		}
+
+		const passwordsMatch = await this.encryptionService.compare(
```
lavandosya, line 38: It's better to start Boolean variables with a prefix that can be answered with "yes" ([B19](https://github.com/BinaryStudioAcademy/quality-criteria/blob/production/src/javascript.md))

```suggestion
		const arePasswordsMatching = await this.encryptionService.compare(
```

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,14 +1,50 @@
...
+
+	public async signIn(
+		userRequestDto: UserSignInRequestDto,
+	): Promise<UserSignUpResponseDto> {
+		const { email, password } = userRequestDto;
+
+		const user = await this.userRepository.findByEmail(email);
+
+		if (!user) {
+			throw new ApplicationError({ message: "Invalid credentials." });
```
lavandosya, line 35: nitpick: Maybe it would be more user-friendly to display "An account with this email does not exist" rather than the generic "Invalid credentials."

apps/backend/src/modules/users/user.entity.ts
```diff
@@ -84,6 +84,14 @@ class UserEntity implements Entity {
 		});
 	}
 
+	public getPasswordSalt(): string {
+		return this.passwordSalt;
+	}
+
+	public getPasswrodHash(): string {
```
lavandosya, line 91:
```suggestion
	public getPasswordHash(): string {
```

lavandosya (changes requested):

apps/backend/src/db/migrations/20250722102254_add_first_name_last_name_to_users.ts
```diff
@@ -16,8 +16,8 @@ function down(knex: Knex): Promise<void> {
 
 function up(knex: Knex): Promise<void> {
 	return knex.schema.alterTable(TABLE_NAME, (table) => {
-		table.string(ColumnName.FIRST_NAME).notNullable();
-		table.string(ColumnName.LAST_NAME).notNullable();
+		table.string(ColumnName.FIRST_NAME).notNullable().defaultTo("");
+		table.string(ColumnName.LAST_NAME).notNullable().defaultTo("");
```
liza-veis, line 20: As discussed,
1. No modification of old migrations - this is very important, so pay attention to this. You may spend a huge amount of time troubleshooting this in a real project, and it can easily lead to a prod crash, just because your backend expected new values after the release, but the migration didn't apply.
2. These fields are required, we don't need default value for them, because we shouldn't have an ability to insert empty values

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -61,6 +63,30 @@ class AuthController extends BaseController {
...
+	}
+
+	public async signIn(
+		options: APIHandlerOptions<{
+			body: UserSignInRequestDto;
+		}>,
+	): Promise<APIHandlerResponse<UserSignUpResponseDto>> {
+		const { body } = options;
+
+		const user = await this.authService.signIn(body);
```
liza-veis, line 84: Since #18 was merged, we need to return the token too, similarly to signUp

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -61,6 +63,30 @@ class AuthController extends BaseController {
...
+				body: userSignInValidationSchema,
+			},
+		});
+	}
+
+	public async signIn(
+		options: APIHandlerOptions<{
+			body: UserSignInRequestDto;
+		}>,
+	): Promise<APIHandlerResponse<UserSignUpResponseDto>> {
```
liza-veis, line 81: Let's create a separate type here even if it fully repeats sign up
```suggestion
	): Promise<APIHandlerResponse<UserSignInResponseDto>> {
```

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,14 +1,52 @@
...
+
+	public async signIn(
+		userRequestDto: UserSignInRequestDto,
+	): Promise<UserSignUpResponseDto> {
+		const { email, password } = userRequestDto;
+
+		const user = await this.userRepository.findByEmail(email);
+
+		if (!user) {
+			throw new ApplicationError({
```
liza-veis, line 35: We are using error per service, so you need to create AuthorizationError that extends HttpError and use it here, plus let's create a enum for messages - AuthorizationExceptionMessage. Similarly to UserError and UserExceptionMessage

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,14 +1,52 @@
...
+				message: "An account with this email does not exist.",
```
liza-veis, line 36: Security tip - do not expose to users whether it's an email or a password that is wrong, just a common message that credentials are invalid

packages/shared/src/modules/users/libs/validation-schemas/user-sign-in.validation-schema.ts
```diff
@@ -0,0 +1,20 @@
...
+	email: z
+		.email({
+			error: UserValidationMessage.EMAIL_WRONG,
+		})
+		.trim()
+		.min(UserValidationRule.EMAIL_MINIMUM_LENGTH, {
+			error: UserValidationMessage.EMAIL_REQUIRED,
+		}),
+	password: z.string().trim().min(UserValidationRule.PASSWORD_MINIMUM_LENGTH, {
+		error: UserValidationMessage.PASSWORD_MINIMUM_LENGTH,
```
liza-veis, line 16: let's add maximum limitation to somewhere around 64 chars

liza-veis (changes requested):

apps/backend/src/modules/auth/auth.service.ts
liza-veis: `backend/src/modules/auth/libs/exceptions/auth.exception.ts`

my-derkach: I just implemented it and push it

packages/shared/src/modules/users/libs/enums/user-validation-rule.enum.ts
```diff
@@ -2,6 +2,7 @@ const UserValidationRule = {
 	EMAIL_MINIMUM_LENGTH: 1,
 	FIRST_NAME_MINIMUM_LENGTH: 2,
 	LAST_NAME_MINIMUM_LENGTH: 2,
+	MAX_LENGTH: 64,
```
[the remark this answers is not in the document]
my-derkach: I use max length for email also, so maybe I will create two different for them EMAIL_MAX_LENGTH and PASSWORD_MAX_LENGTH

my-derkach:
So in total, I change toObject() method inside UserEntity, for now it takes optional argument (includePassword), and if it is true we retrun base object with passwords. Based on it I expand UserGetByIdItemResponseDto with optional passwords. And finally just update auth service, namely signIn

liza-veis (approved):

packages/shared/src/modules/users/libs/types/user-get-all-item-response-dto.type.ts
```diff
@@ -3,6 +3,8 @@ type UserGetByIdItemResponseDto = {
 	firstName: string;
 	id: number;
 	lastName: string;
+	passwordHash?: string;
+	passwordSalt?: string;
```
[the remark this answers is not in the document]
my-derkach: Fix that)

kalinkaaaa14 (approved):

liza-veis (approved):

PR #74, opened 2025-07-24
feat(frontend): landing and application separation ss-53

DenisDemirov:
- Application separated on two parts
- Added hero section in landing

зображення


apps/frontend/src/assets/css/variables.css
```diff
@@ -38,4 +38,5 @@
 	--font-size-large: 2.25rem;
 	--font-weight-regular: 400;
 	--font-weight-semibold: 600;
+	--font-weight-bold: 700;
```
liza-veis, line 41: We need to add this font file and font-face declaration too. It's unfortunate we have to add it only for the landing h1 title, but it doesn't look nice with semi-bold, so let it be...

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -14,6 +14,7 @@ import {
 } from "~/libs/hooks/hooks.js";
 import { actions as userActions } from "~/modules/users/users.js";
 
+import { Landing } from "../landing/landing.js";
```
liza-veis, line 17: Landing should be separate from our app

apps/frontend/src/libs/components/private-route/private-route.tsx
liza-veis, line 1: We will add protected routing later, it will depend on user authentication, but it has nothing to do with this ticket, so let's remove this

apps/frontend/src/index.tsx
```diff
@@ -31,6 +32,10 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
 						element: <App />,
 						path: AppRoute.ROOT,
 					},
+					{
+						element: <PrivateRoute />,
+						path: AppRoute.APP,
+					},
```
liza-veis, line 38: it should look like this
```suggestion
element: <App />,
path: AppRoute.APP,
},
					{
						element: <Landing />,
						path: AppRoute.ROOT,
					},
```

apps/frontend/src/libs/components/landing/landing.tsx
```diff
@@ -0,0 +1,26 @@
+import styles from "./styles.module.css";
+
+const Landing = (): React.JSX.Element => {
+	return (
+		<main className={styles["landing"]}>
+			<div className={styles["landing-header"]}>
```
liza-veis, line 6: We will create a separate component per section - `landing/libs/components/hero-section/hero-section.tsx`. And it's better to rename header to hero here, because header is confusing

apps/frontend/src/libs/components/landing/landing.tsx
```diff
@@ -0,0 +1,26 @@
...
+				<h1 className={styles["landing-title"]}>
+					Discover Your Next Adventure
+				</h1>
+
+				<h4 className={styles["landing-description"]}>
```
liza-veis, line 11: It's just a paragraph from seo perspective. Tags value from seo perspective, and the actual style name on the design may not always match.
```suggestion
				<p className={styles["landing-description"]}>
```

apps/frontend/src/libs/components/landing/styles.module.css
```diff
@@ -0,0 +1,38 @@
...
+	display: flex;
+	flex-direction: column;
+	gap: 16px;
+	width: 100%;
+	margin: 100px 0;
+}
+
+.landing-title {
+	margin-bottom: 0;
+	font-size: 3.5rem;
```
liza-veis, line 19: Don't we have variable for this? if not let's create it

liza-veis (changes requested):

apps/frontend/src/libs/enums/app-route.enum.ts
```diff
@@ -1,7 +1,8 @@
 const AppRoute = {
+	APP: "/app",
 	ROOT: "/",
-	SIGN_IN: "/sign-in",
-	SIGN_UP: "/sign-up",
+	SIGN_IN: "sign-in",
+	SIGN_UP: "sign-up",
```
liza-veis, line 5: hm, why this change is needed
```suggestion
	SIGN_IN: "/sign-in",
	SIGN_UP: "/sign-up",
```

apps/frontend/src/index.tsx
```diff
@@ -12,6 +12,8 @@ import { store } from "~/libs/modules/store/store.js";
 import { Auth } from "~/pages/auth/auth.jsx";
 import { NotFound } from "~/pages/not-found/not-found.jsx";
 
+import { Landing } from "./libs/components/landing/libs/components/hero-section/hero-section.js";
```
liza-veis, line 15: You need to have a landing component 
```suggestion
import { Landing } from "./libs/components/landing/landing.jsx";
```

and this landing component, will combine multiple sections on it as sub components

apps/frontend/src/libs/components/landing/libs/components/hero-section/hero-section.tsx
```diff
@@ -0,0 +1,26 @@
+import styles from "./styles.module.css";
+
+const Landing = (): React.JSX.Element => {
+	return (
+		<main className={styles["landing"]}>
+			<div className={styles["landing-hero"]}>
+				<h1 className={styles["landing-title"]}>
```
liza-veis, line 7: As mentioned above
```tsx
const Landing = (): React.JSX.Element => {
	return (
		<main className={styles["landing"]}> // this should go to landing/landing.tsx
			<section className={styles["section"]}> // this will remain here and the component name will be HeroSection
				<h1 className={styles["title"]}>
```

liza-veis (changes requested):

apps/frontend/src/libs/components/landing/libs/components/hero-section/hero-section.tsx
h4vefundieyoung: you do. considering it also would be nice to move things into page folder.

apps/frontend/src/index.tsx
```diff
@@ -12,6 +12,8 @@ import { store } from "~/libs/modules/store/store.js";
 import { Auth } from "~/pages/auth/auth.jsx";
 import { NotFound } from "~/pages/not-found/not-found.jsx";
 
+import { Landing } from "./libs/components/landing/landing.jsx";
```
kalinkaaaa14, line 15: In src/index we should place imports pages that are visible on different routes. So similar to auth and NotFound, there should be separate page created for landing

DenisDemirov: Hmm... That what i already done ? Or i understand it wrong ?
зображення

kalinkaaaa14: You should have like this
pages/landing.tsx
pages/libs/components/hero-section

In src/libs/components we should put components which will be reusable on the project, for example buttons, inputs, some styled cards etc. Some specific components to the pages should be placed in pages/libs/components

apps/frontend/src/libs/enums/app-route.enum.ts
liza-veis: I see, I will take a look at this

apps/frontend/src/libs/components/landing/libs/components/hero-section/hero-section.tsx
liza-veis: yep, that's right

apps/frontend/src/index.tsx
kalinkaaaa14: This is not resolved as I see, changes need to be made on this comment too

DenisDemirov: @liza-veis Ok, it fixed

apps/frontend/src/index.tsx
```diff
@@ -12,6 +12,8 @@ import { store } from "~/libs/modules/store/store.js";
 import { Auth } from "~/pages/auth/auth.jsx";
 import { NotFound } from "~/pages/not-found/not-found.jsx";
 
+import { Landing } from "./pages/landing.js";
```
kalinkaaaa14, line 15:
```suggestion
import { Landing } from "./pages/landing/landing.jsx";
```

liza-veis (changes requested):

liza-veis (approved):

PR #82, opened 2025-07-24
feat(backend/shared): add protected routing ss-23

h4vefundieyoung:
- Backend protected routing has been implemented via custom fastify plugin.
- Authorized user entity injected directly into request object.
- Authorized user entity injected into controller handler.
- UserGetAllItemResponseDto interface renamed to UserGetByIdItemResponseDto

https://github.com/user-attachments/assets/116a8713-d576-407a-869e-11ff96ccc79e


apps/backend/src/libs/modules/plugins/auth/base-auth.plugin.ts
```diff
@@ -0,0 +1,77 @@
...
+type TokenPayload = {
+	payload: {
+		userId: number;
+	};
+};
+
+class BaseAuthPlugin implements AuthPluginApi {
+	private readonly jwtService: BaseToken;
+	private readonly userService: UserService;
+	private readonly whiteListRoutes: string[] = [];
```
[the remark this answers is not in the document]
h4vefundieyoung: whatcha think?

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -28,6 +28,12 @@ class UserRepository implements Repository {
 		return UserEntity.initialize(user);
 	}
 
+	public async find(id: number): Promise<null | UserEntity> {
+		const user = await this.userModel.query().where("id", id).first();
```
kalinkaaaa14, line 32:
```suggestion
		const user = await this.userModel.query().findById(id);
```

apps/backend/src/libs/modules/plugins/auth/base-auth.plugin.ts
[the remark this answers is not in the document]
kalinkaaaa14: I think that is could be done easier, you need something like this

```
type Options = {
  routesWhiteList: WhiteRoute[];
};

const auth: FastifyPluginAsync<Options> = async (fastify, opts) => {
  fastify.decorateRequest('user', null);

  fastify.addHook('onRequest', async (request, reply) => {
    // set request.user here if the route is not in the whitelist
  });
};
const authorization = fastifyPlugin(auth);

```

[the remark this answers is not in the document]
kalinkaaaa14: WhiteRoute could be something like
```
type WhiteRoute = {
  route: string;
  methods: HTTPMethod[];
};
```

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -195,6 +199,10 @@ class BaseServerApplication implements ServerApplication {
 		this.addRoutes(routers);
 	}
 
+	private async initRoutesProtection(): Promise<void> {
+		await this.app.register(fastifyPlugin(authPlugin.plugin.bind(authPlugin)));
+	}
```
kalinkaaaa14, line 204:
```suggestion
	private async initPlugins(): Promise<void> {
		await this.app.register(authPlugin, <throw whitelists here>);
	}
```

apps/backend/src/libs/modules/plugins/auth/base-auth.plugin.ts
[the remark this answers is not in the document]
h4vefundieyoung: > I think that is could be done easier, you need something like this
> 
> ```
> type Options = {
>   routesWhiteList: WhiteRoute[];
> };
> 
> const auth: FastifyPluginAsync<Options> = async (fastify, opts) => {
>   fastify.decorateRequest('user', null);
> 
>   fastify.addHook('onRequest', async (request, reply) => {
>     // set request.user here if the route is not in the whitelist
>   });
> };
> const authorization = fastifyPlugin(auth);
> ```

well, yeah, i'm agree that it looks much simpler, but whole the backend app - is a bunch of classes, should we really describe these plugins as a functions?

liza-veis (changes requested):

apps/backend/src/libs/modules/plugins/auth/base-auth.plugin.ts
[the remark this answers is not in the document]
liza-veis: Thanks for asking and trying to understand the architecture. As you noticed, in our app, we have classes for the usual layers in a REST application (controller, service, repository), classes to work with the data (entities/models), and orchestration classes (ServerApplication) or classes for specific modules (e.g., Logger). This may look like a distant version of OOP, but not to the full extent, obviously. For the plugins, we mostly prefer to use a simplified version similar to hooks or middlewares, and you are asking a good question - why. The main reason is that for plugins/hooks/middlewares, we prefer a functional approach, and they are being applied in Fastify using a functional approach (you pass your middleware functions as an argument to a specific binding function in Fastify). So, in short, our project doesn't stick directly to OOP or functional approach, but rather uses common patterns in solving problems and whatever usually works better. For middlewares (plugins/hooks are actually the same thing here), we usually don't have a lot of logic in them, but may have a lot of them as helpers-like functions, and will apply them in a functional style. It means your approach is not wrong, and it can make sense in a project where this is the expected architecture. For our project for such entities like hooks/middlewares/plugins, we will use a functional approach as more conventional in the field, and in case it will require to work with a complex logic, we will move this logic out to a separate module/service/helpers and re-use this service in the middleware to keep it clean.

If you have any additional questions, I would be happy to reply 😎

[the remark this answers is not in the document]
h4vefundieyoung: > Thanks for asking and trying to understand the architecture. As you noticed, in our app, we have classes for the usual layers in a REST application (controller, service, repository), classes to work with the data (entities/models), and orchestration classes (ServerApplication) or classes for specific modules (e.g., Logger). This may look like a distant version of OOP, but not to the full extent, obviously. For the plugins, we mostly prefer to use a simplified version similar to hooks or middlewares, and you are asking a good question - why. The main reason is that for plugins/hooks/middlewares, we prefer a functional approach, and they are being applied in Fastify using a functional approach (you pass your middleware functions as an argument to a specific binding function in Fastify). So, in short, our project doesn't stick directly to OOP or functional approach, but rather uses common patterns in solving problems and whatever usually works better. For middlewares (plugins/hooks are actually the same thing here), we usually don't have a lot of logic in them, but may have a lot of them as helpers-like functions, and will apply them in a functional style. It means your approach is not wrong, and it can make sense in a project where this is the expected architecture. For our project for such entities like hooks/middlewares/plugins, we will use a functional approach as more conventional in the field, and in case it will require to work with a complex logic, we will move this logic out to a separate module/service/helpers and re-use this service in the middleware to keep it clean.
> 
> If you have any additional questions, I would be happy to reply 😎

well, I got the main idea, thanks for such detailed answer, I'll refactor it with next commit 😸

apps/backend/src/modules/users/user.service.ts
```diff
@@ -45,6 +45,12 @@ class UserService implements Service {
 		return item.toObject();
 	}
 
+	public async find(id: number): Promise<null | UserGetAllItemResponseDto> {
```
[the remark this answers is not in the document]
kalinkaaaa14: I think that it's better to do it now, until we do not forget about it and there is no large amount of changes in main regarding this type. And if anyone is using the old version of this type, he should resolve it during merging conflicts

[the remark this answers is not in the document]
h4vefundieyoung: > I think that it's better to do it now, until we do not forget about it and there is no large amount of changes in main regarding this type. And if anyone is using the old version of this type, he should resolve it during merging conflicts

well, it's done, but i believe we will face some conflicts soon

apps/backend/src/libs/modules/plugins/auth.plugin.ts
```diff
@@ -0,0 +1,73 @@
...
+		);
+
+		if (isWhiteListRoute) {
+			return;
+		}
+
+		if (!headers.authorization) {
+			sendUnathorized(reply);
+
+			return;
```
kalinkaaaa14, line 48:
```
throw new AuthError({
				message: AuthExceptionMessage.UNAUTHORIZED_REQUEST,
				status: HTTPCode.UNAUTHORIZED,
			});
```

apps/backend/src/libs/modules/plugins/auth.plugin.ts
```diff
@@ -0,0 +1,73 @@
...
+		}
+
+		try {
+			const tokenIndex = 1;
+			const token = headers.authorization.split(" ")[tokenIndex] as string;
```
kalinkaaaa14, line 53: `const [, authToken] = headers.authorization.split(' ') ?? [];
`

apps/backend/src/libs/modules/plugins/auth.plugin.ts
```diff
@@ -0,0 +1,73 @@
...
+			const { userId } = await tokenService.verify<TokenPayload>(token);
+			const user = await userService.findById(userId);
+
+			if (user) {
+				request.user = user;
+			} else {
+				sendUnathorized(reply);
+			}
```
kalinkaaaa14, line 61:
```
if(!user){
throw new AuthError({
				message: AuthExceptionMessage.UNAUTHORIZED_REQUEST,
				status: HTTPCode.UNAUTHORIZED,
			});
}

request.user = user;
```

apps/backend/src/libs/modules/plugins/auth.plugin.ts
```diff
@@ -0,0 +1,73 @@
...
+type TokenPayload = {
+	userId: number;
+};
+
+type WhiteRoute = {
+	method: HTTPMethod;
+	path: string;
+};
+
+const authPlugin = (
```
kalinkaaaa14, line 27:
```suggestion
const auth = (
```

apps/backend/src/libs/modules/plugins/auth.plugin.ts
```diff
@@ -0,0 +1,73 @@
...
+		} catch {
+			sendUnathorized(reply);
+		}
+	};
+
+	app.decorate("user", null);
+	app.addHook("onRequest", requestHandler);
+};
+
+const fp = fastifyPlugin<PluginOptions>(authPlugin);
```
kalinkaaaa14, line 71:
```suggestion
const authPlugin = fastifyPlugin<PluginOptions>(authPlugin);
```

h4vefundieyoung:
I asked about access level before, like a admin access level, initially I wanted to implement abstract layer for different auth strategies, but as I remember instructions was like a - such thing may be controlled from the controller handler directly, according to that point I described the auth helper. If it's definitely won't be need I can remove it

kalinkaaaa14:
> such thing may be controlled from the controller handler directly,
I totally agree with that, and from my perspective it means that we do not need neither abstract layers, nor helpers, as we can simply throw appropriate errors directly where we need to

But let's sync with @liza-veis or @lavandosya what they are thinking about this

apps/backend/src/libs/modules/plugins/auth.plugin.ts
h4vefundieyoung: > `const [, authToken] = headers.authorization.split(' ') ?? []; `

what the purpose of this nullish coalescing? split function will return an array in any case

h4vefundieyoung:
> I also support @kalinkaaaa14 point. This way, we are going to have a single approach in handling error cases in our app - throwing custom errors which will be later handled in a single middleware and formatted to an appropriate response and sent to the frontend

well, then I believe that's the decision

apps/backend/src/modules/users/user.service.ts
```diff
@@ -45,21 +45,25 @@ class UserService implements Service {
 		return item.toObject();
 	}
 
-	public async findAll(): Promise<CollectionResult<UserGetAllItemResponseDto>> {
+	public async findAll(): Promise<CollectionResult<UserItemResponseDto>> {
```
liza-veis, line 48: Let' rename it to the convention we have; sometimes, we will reuse those types for other methods too, like in this case (we have base type, get by id; get all for now matches, so we use the same time, but maybe we will create another method in the future). By our style guides we need to name it including the action 
```suggestion
	public async findAll(): Promise<CollectionResult<UserGetByIdItemResponseDto>> {
```

liza-veis (changes requested):
Looks good overall, a couple of comments 😎

apps/backend/src/modules/users/user.service.ts
h4vefundieyoung: it's already have been renamed

apps/backend/src/libs/modules/server-application/libs/constants/constants.ts
```diff
@@ -0,0 +1,14 @@
+import { APIPath, AuthApiPath } from "~/libs/enums/enums.js";
+
+const WHITE_ROUTES = [
+	{
+		method: "POST" as const,
+		path: APIPath.AUTH + AuthApiPath.SIGN_IN,
```
kalinkaaaa14, line 6:
```suggestion
		path: `${APIPath.AUTH}${AuthApiPath.SIGN_IN}`,
```

apps/backend/src/libs/modules/server-application/libs/constants/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+	},
+	{
+		method: "POST" as const,
+		path: APIPath.AUTH + AuthApiPath.SIGN_UP,
```
kalinkaaaa14, line 10:
```suggestion
		path: `${APIPath.AUTH}${AuthApiPath.SIGN_UP}`,
```

apps/backend/src/libs/modules/server-application/libs/constants/constants.ts
```diff
@@ -0,0 +1,14 @@
+import { APIPath, AuthApiPath } from "~/libs/enums/enums.js";
+
+const WHITE_ROUTES = [
```
kalinkaaaa14, line 3:
```suggestion
const WHITE_ROUTES: WhiteRoute[] = [
```

apps/backend/src/libs/modules/plugins/auth/auth.plugin.ts
```diff
@@ -0,0 +1,59 @@
...
+
+		const [, token] = headers.authorization.split(" ");
+		const { userId } = await tokenService.verify<TokenPayload>(token as string);
+		const user = await userService.findById(userId);
+
+		if (user) {
+			request.user = user;
+		} else {
+			throw new AuthError();
+		}
```
kalinkaaaa14, line 50: You don't need else block here

```suggestion
		if (!user) {
		    throw new AuthError();
		}
		
			request.user = user; 
```

kalinkaaaa14 (approved):

apps/backend/src/libs/modules/server-application/libs/constants/constants.ts
```diff
@@ -0,0 +1,16 @@
+import { APIPath, AuthApiPath } from "~/libs/enums/enums.js";
+
+import { type WhiteRoute } from "../types/types.js";
+
+const WHITE_ROUTES: WhiteRoute[] = [
+	{
+		method: "POST" as const,
```
kalinkaaaa14, line 7: doesn't it work without as const?
```suggestion
		method: "POST",
```

h4vefundieyoung: @kalinkaaaa14 I believe it does, forget to remove after I've added type directly to variable.
I have a question. you asked me to add app.decorate("user", null) - do we really need it? still can't get the purpose, I mean we attach user entity to the request, not to the app.

h4vefundieyoung: > doesn't it work without as const?

I believe it does. just forgot to remove after I've added the type directly to variable.

kalinkaaaa14 (approved):

liza-veis (approved):

PR #86, opened 2025-07-25
feat(backend): eliminate unnecessary payload nesting when creating JWT token ss-85

### LINKED ISSUE #85: feat: Eliminate unnecessary payload nesting when creating JWT token
### What feature?

Current implementation results in the following token payload:
```js
{
 payload: {
  payload: {
   userId
  }
 }
}
```

Expected payload:

```js
{
 payload: {
   userId
 }
}
```

### Screenshots

_No response_

mxm-kotyk:
Resolves #85


liza-veis (approved):

apps/backend/src/libs/modules/token/base-token.module.test.ts
```diff
@@ -30,10 +30,11 @@ describe("Token service", () => {
...
 		});
 
-		const token = await tokenService.create(userId);
+		const token = await tokenService.create({ userId });
 
-		const { payload } = await tokenService.verify(token);
-		assert.equal(payload, userId);
+		const payload = await tokenService.verify(token);
+
+		assert.equal(payload["userId"], userId);
```
h4vefundieyoung, line 37: would be great to replace it with prop name with id value

mxm-kotyk: Done

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -24,7 +24,7 @@ class AuthService {
 		userRequestDto: UserSignUpRequestDto,
 	): Promise<UserSignUpResponseDto> {
 		const { email, id } = await this.userService.create(userRequestDto);
-		const token = await this.tokenService.create({ userId: id });
+		const token = await this.tokenService.create({ id });
```
liza-veis, line 27: I believe @h4vefundieyoung comment was regarding the removal of user_id from base-token.module.test.ts, as it could accept any data across the app. Please correct me if I'm mistaken. For the actual token, we need to include meaningful data that may vary depending on the context to help identify the user or resource. So for authorization token, let's keep userId as it is

apps/backend/src/libs/modules/token/base-token.module.test.ts
```diff
@@ -30,10 +30,11 @@ describe("Token service", () => {
 			tokenExpirationTime: TOKEN_EXPIRATION,
 		});
 
-		const token = await tokenService.create(userId);
+		const token = await tokenService.create({ id: userId });
```
liza-veis, line 33:
```suggestion
		const token = await tokenService.create({ id: resourceId });
```

liza-veis (changes requested):

h4vefundieyoung:
well, ye, one more naming thing, all these names - i believe we should rename things with "jwt" pattern. like a jwt-service, jwt-controller, jwt-whatever, just to clarify what the token is. there may be plenty of different token-things in the app.

apps/backend/src/modules/auth/auth.service.ts
h4vefundieyoung: > I believe @h4vefundieyoung comment was regarding the removal of user_id from base-token.module.test.ts, as it could accept any data across the app. Please correct me if I'm mistaken. For the actual token, we need to include meaningful data that may vary depending on the context to help identify the user or resource. So for authorization token, let's keep userId as it is

well, not really, i was talking about common pattern in the app when we use "userId" prop instead of "id", that leads to unnecesary interface declarataion, just makes sense to stay close to the common pattern in this case as for me.

h4vefundieyoung: and the case you've describe with testing - is just another case that should be considered

mxm-kotyk:
> well, ye, one more naming thing, all these names - i believe we should rename things with "jwt" pattern. like a jwt-service, jwt-controller, jwt-whatever, just to clarify what the token is. there may be plenty of different token-things in the app.

So how should it look like?
Something like this:
```
class BaseJwt implements Jwt
```
for entities, and like this:
```
jwt/
├─ libs/
│  └─ types/
│     ├─ jwt-payload.type.ts
│     ├─ jwt.type.ts
│     └─ jwt.ts
├─ base-jwt.module.test.ts
├─ base-jwt.module.ts
└─ jwt.ts
```
for files and folders?

liza-veis:
> Well, ye, one more naming thing, all these names - i believe we should rename things with "jwt" pattern. like a jwt-service, jwt-controller, jwt-whatever, just to clarify what the token is. there may be plenty of different token-things in the app.

This is a reasonable architectural approach that may fit well in medium-sized and large-scale applications. And I also want to point out that it would be correct from a usual OOP perspective, because we would have our single abstract class (or interface) and a specific implementation of it (JWT class). However, every layer of abstraction has its cost, and our project is not that large and intricate to consider possible multiple implementations, at least for now. 

So, as you may see across the project, we are mostly sticking to a singleton architecture pattern for modules, and this pattern implies that we will only have a single instance/implementation that will be used in our app. Obviously, as the project extends, we may leverage other more complex architectural patterns too if this would be needed. But for now, we are having a single implementation, and there is no need for the application to know details of that implementation. So, in case of the singleton pattern, it's a bit vice-versa, and we want to abstract the application (through an exposed unified interface) from the implementation details. For example, to consider for the cases like @mxm-kotyk  had mentioned:

> What if the future project developers decide to replace JWT with another technology that encrypts/decrypts data into token and not relies on JSON? Like PASETO?

In case we are not planning to change the specific module architecture from the singleton pattern, we would prefer to update its internal implementation details rather than creating a separate class. And it's totally valid, if such a case appears, and a team may consider that the module should be rewritten to another type of architecture (in this case, the approach with separate implementation will work fine). But what's important to understand here is that this will be a shift in architectural pattern (more significant for the system, that it may look like at first sight) and would need to be evaluated against its trade-offs and advantages carefully 😎 This part was a bit off topic, but I thought this way of thinking about systems could be interesting to some of you.

apps/backend/src/modules/auth/auth.service.ts
h4vefundieyoung: > > well, not really, i was talking about common pattern in the app. when we use "userId" prop instead of "id" that leads to unnecesary interface declarataion, just makes sense to stay close to the common pattern in this case as for me.
> 
> Oh, I see your point. Then I would say that for tokens, having a more specific payload seems to be more beneficial to me. There are a few reasons for this:
> 
> 1. Over time, you will quickly forget what this id actually meant in terms of token and token is very isolated from the application and any module we have there, so we won't have any context and we will need to specifically remember what it meant, or to go through the flow and logic to find this out which can become tricky when project grows.
> 2. It's not that unusual to have multiple tokens in the app for different purposes (e.g., password reset). As our application will handle multiple tokens, it will become crucial to differentiate between them to avoid mixing different tokens.
> 3. As @mxm-kotyk, it's also pretty common to pass additional data - everything that we need to identify the operation. For example, project ID, operation type, related entity type, or anything that is needed for the context, because the purpose of that payload is to keep the needed context to identify and approve the operation later.
> 4. More specific interfaces also allow us to better cover it with typing and to avoid passing what we weren't supposed to. From the code perspective, it's relatively easy to pass not the needed entity id, but the id of anything else, and it would be really easy to miss this mistake for reviewers, too

well, yea, i understand that payload and a general context can be different a lot, all i'm trying to say is that we already have lots of described interfaces in the app, where we're using "id" prop, so probably it would be nice to reuse them in generic type for jwt verify method and not to describe additional interfaces for both apps just because we passed userId instead of id into payload.

h4vefundieyoung: > @h4vefundieyoung @liza-veis
> 
> So we all agree to use `{ userId: id }` as argument for token creation in auth.service? And I change `tokenService.create({ id: userId })` in base-token.module.test.ts to `tokenService.create({ id: resourceId })` as was proposed earlier.

i believe that i've described some "over optimization" stuff, and the case's definitely not a big deal, right now i just need this branch to be merged

liza-veis (approved):

PR #89, opened 2025-07-25
feat(frontend): enable PWA for frontend ss-68

### LINKED ISSUE #68: feat: Enable PWA for the frontend application
### What feature?

https://vite-pwa-org.netlify.app/

- [ ] Enable and set up PWA
- [ ] Yser should be able to download the web app

### Screenshots

_No response_

ZakZh:
Feature: Enable PWA functionality

- What's implemented:
Complete PWA setup using vite-plugin-pwa
Service Worker registration with auto-updates
App manifest with proper icons and metadata
Installation capability - users can download/install the web app

Closes #68

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,6 +1,6 @@
 {
-	".": "1.9.3",
-	"apps/backend": "1.6.3",
+	".": "1.10.0",
+	"apps/backend": "1.7.0",
 	"apps/frontend": "1.9.3",
-	"packages/shared": "1.7.1"
+	"packages/shared": "1.8.0"
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [1.10.0](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/compare/v1.9.3...v1.10.0) (2025-08-04)
+
+
+### Features
+
+* **backend/shared:** add postgis ss-112 ([#121](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/issues/121)) ([d6b4932](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/commit/d6b49329a426a1ca1bb48a22757d547edf86dbe6))
+
 ## [1.9.3](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/compare/v1.9.2...v1.9.3) (2025-08-01)
 
 
--- a/apps/backend/CHANGELOG.md
+++ b/apps/backend/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [1.7.0](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/compare/backend-v1.6.3...backend-v1.7.0) (2025-08-04)
+
+
+### Features
+
+* **backend/shared:** add postgis ss-112 ([#121](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/issues/121)) ([d6b4932](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/commit/d6b49329a426a1ca1bb48a22757d547edf86dbe6))
+
 ## [1.6.3](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/compare/backend-v1.6.2...backend-v1.6.3) (2025-08-01)
 
 
--- a/apps/backend/package.json
+++ b/apps/backend/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "@smartscapes/backend",
 	"type": "module",
-	"version": "1.6.3",
+	"version": "1.7.0",
 	"engines": {
 		"node": ">=22.10.0 <23.0.0",
 		"npm": "10.x.x"
--- a/apps/backend/src/db/migrations/20250729155425_add_postgis_extension.ts
+++ b/apps/backend/src/db/migrations/20250729155425_add_postgis_extension.ts
@@ -1,0 +1,11 @@
+import { type Knex } from "knex";
+
+async function down(knex: Knex): Promise<void> {
+	await knex.raw("DROP EXTENSION IF EXISTS postgis");
+}
+
+async function up(knex: Knex): Promise<void> {
+	await knex.raw("CREATE EXTENSION IF NOT EXISTS postgis");
+}
+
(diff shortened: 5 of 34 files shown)


apps/frontend/src/libs/modules/pwa/pwa.module.ts
```diff
@@ -0,0 +1,53 @@
+// eslint-disable-next-line import/no-unresolved
```
ZakZh, line 1: I had to exclude it, since the module doesn’t actually exist and the linter wouldn’t let it through.
After some searching, I found options to exclude it globally in the linter config for  "virtual:", or to exclude it locally. Other approaches I found didn’t pass the linter either.
Знімок екрана 2025-07-25 184501

apps/frontend/src/libs/modules/pwa/libs/enums/pwa-config.enum.ts
```diff
@@ -0,0 +1,10 @@
+const MILLISECONDS_IN_SECOND = 1000;
+const SECONDS_IN_MINUTE = 60;
+const MINUTES_IN_HOUR = 60;
+
+const PWAConfig = {
+	UPDATE_INTERVAL_MS:
+		MINUTES_IN_HOUR * SECONDS_IN_MINUTE * MILLISECONDS_IN_SECOND, // 1 hour
```
lavandosya, line 7: [B29](https://github.com/BinaryStudioAcademy/quality-criteria/blob/production/src/javascript.md): Constants of the same category should be grouped using enums for better structure and readability

apps/frontend/src/libs/modules/pwa/pwa.module.ts
h4vefundieyoung: most likely should be resolved through config

apps/frontend/src/libs/modules/pwa/libs/enums/pwa-config.enum.ts
```diff
@@ -0,0 +1,10 @@
+const MILLISECONDS_IN_SECOND = 1000;
+const SECONDS_IN_MINUTE = 60;
+const MINUTES_IN_HOUR = 60;
```
h4vefundieyoung, line 3: i'd say it just looks redundant.

ZakZh: Agreed, but it’s more like about “the right way” than an actual need.

ZakZh: For now, I’ll use it locally, but if we decide to go with it, I’ll rework it.
Anyway, thanks!

apps/frontend/src/libs/modules/pwa/libs/enums/update-interval.enum.ts
```diff
@@ -0,0 +1,14 @@
...
+	MINUTES_IN_HOUR: 60,
+	SECONDS_IN_MINUTE: 60,
+} as const;
+
+const UpdateInterval = {
+	DEFAULT:
+		TimeConstants.MINUTES_IN_HOUR *
+		TimeConstants.SECONDS_IN_MINUTE *
+		TimeConstants.MILLISECONDS_IN_SECOND,
+} as const;
```
liza-veis, line 12: Time constant is not an enum by its logic, more like a set of mapped constants, so let's move it to frontend/libs/constants level. For the update interval, it should be in local constants and let's simplify it to singular value, I beleive we won't need to add other options
```suggestion
const TIME_UNIT = {
  MILLISECONDS_IN_SECOND: 1000,
  SECONDS_IN_MINUTE: 60,
  MINUTES_IN_HOUR: 60,
} as const;

const UPDATE_INTERVAL = TimeConstant.MINUTES_IN_HOUR *
    TimeConstant.SECONDS_IN_MINUTE *
    TimeConstant.MILLISECONDS_IN_SECOND;
```

apps/frontend/src/libs/modules/pwa/libs/types/types.ts
```diff
@@ -0,0 +1,9 @@
+type PWARegistrationConfig = {
+	immediate?: boolean;
```
liza-veis, line 2: Boolean vars should start with `is`, `has`, `should` etc, something you can reply with true or false
```suggestion
	isImmediate?: boolean;
```

apps/frontend/src/libs/modules/pwa/pwa.module.ts
```diff
@@ -0,0 +1,52 @@
...
+			immediate = true,
+			onNeedRefresh,
+			onRegistered,
+			updateIntervalMs = this.updateInterval,
+		} = config;
+
+		registerSW({
+			immediate,
+			onNeedRefresh: (): void => {
+				if (import.meta.env.DEV) {
```
liza-veis, line 27: let's pass it through props to the constructor, by the way, what is this variable about? Is it checking dev mode or what? I don't see any changes in .env.example, but for the environment we already have a specified variable in config - ENVIRONMENT and an enum - AppEnvironment with possible values

ZakZh: Regarding that variable — it’s part of the actual PWA implementation. According to the library’s documentation (and as we agreed), we use two mechanisms for automatic updates: а periodic check (every hour) and the standard registerType: "autoUpdate" mode, which depends on the browser’s behavior and its handling of the manifest and service worker. 
The onNeedRefresh callback, in this case, is triggered when an update is available — and in development mode (import.meta.env.DEV), it forces an immediate refresh. That’s useful during development but wouldn’t be the standard behavior in production.

ZakZh: Yes, i'll do that ASAP

apps/frontend/src/libs/modules/pwa/pwa.module.ts
```diff
@@ -0,0 +1,44 @@
...
+				}, updateIntervalMs);
+
+				onRegistered?.(registration);
+			},
+		} satisfies RegisterSWOptions);
+	}
+
+	public setUpdateInterval(intervalMs: number): void {
+		this.updateInterval = intervalMs;
+	}
```
liza-veis, line 41: seems like it's not used, so let's remove it
```suggestion
```

liza-veis (commented):
One comment, other than that LGTM

liza-veis (approved):

lavandosya (approved):

PR #90, opened 2025-07-26
feat(backend/shared): add reviews module ss-75

### LINKED ISSUE #75: feat: Add reviews module
### What feature?

Create a separate module on the backend to work with poi/route reviews entity

- [x] Add new table to DB 

```
Table reviews {
  id integer [primary key]
  created_at timestamp
  updated_at timestamp
  user_id integer [not null]
  content text
  likes_count integer
  route_id integer
  poi_id integer
}
```

- [x] Add `/reviews` controller
- [x] Add an ability to create a review through api
- [x] Either route_id or poi_id is required
- [x] Cover new functionality with tests

### Screenshots

_No response_

vldkorin:
Backend:
 - added new table to DB
 - separate reviews module which includes:
   -  exception message enum
   -  review exception
   -  types for request and response dto, validation-schemas for creating review (exported from shared)
   - controller, entity, model, repository, service
   - covered with tests: controller, entity, repository, service

Shared:
- separate reviews module which includes:
   - validation message, rule enums
   - review create validation schema
   - response, request dto types

Also:
- Swagger documentation for controller
- Updated db schema(readme.md)
 
 Closes #75


apps/backend/src/db/migrations/20250726190816_add_reviews_table.ts
```diff
@@ -0,0 +1,39 @@
...
+		table.increments(ColumnName.ID).primary();
+		table
+			.dateTime(ColumnName.CREATED_AT)
+			.notNullable()
+			.defaultTo(knex.fn.now());
+		table
+			.dateTime(ColumnName.UPDATED_AT)
+			.notNullable()
+			.defaultTo(knex.fn.now());
+		table.integer(ColumnName.USER_ID).notNullable();
```
kalinkaaaa14, line 31: It should be reference to users table

apps/backend/src/db/migrations/20250726190816_add_reviews_table.ts
```diff
@@ -0,0 +1,39 @@
...
+		table.text(ColumnName.CONTENT);
+		table.integer(ColumnName.LIKES_COUNT).defaultTo(0);
+		table.integer(ColumnName.ROUTE_ID).nullable();
+		table.integer(ColumnName.POI_ID).nullable();
```
kalinkaaaa14, line 35: If I'm not mistaken, columns are nullable by default. You should mark required fields as not Nullable

kalinkaaaa14: Also, you need to add a constraint on the DB level to check that one of these columns - poi_id / route_id is required, so is not null

packages/shared/src/modules/reviews/libs/validation-schemas/review-create.validation.schema.ts
```diff
@@ -0,0 +1,28 @@
+import { z, type ZodType } from "zod";
+
+import {
+	ReviewValidationMessage,
+	ReviewValidationRule,
+} from "../enums/enums.js";
+import { type ReviewRequestDto } from "../types/types.js";
+
+const reviewCreate: ZodType<ReviewRequestDto> = z
```
kalinkaaaa14, line 9:
```suggestion
const reviewCreate = z
```

packages/shared/src/modules/reviews/libs/types/review-request-dto.type.ts
```diff
@@ -0,0 +1,8 @@
+type ReviewRequestDto = {
+	content: string;
+	poiId: null | number;
+	routeId: null | number;
+	userId: number;
```
kalinkaaaa14, line 5: User id should be taken from request.user field, it should not be passed to the request body. For now we can leave it, but we will definitely need to refactor it after ticket about route protection on backend will be merged

apps/backend/src/modules/reviews/review.service.ts
```diff
@@ -0,0 +1,57 @@
...
+				status: HTTPCode.BAD_REQUEST,
+			});
+		}
+
+		if (payload.routeId && payload.poiId) {
+			throw new ReviewError({
+				message: ReviewExceptionMessage.ROUTE_OR_POI_REQUIRED,
+				status: HTTPCode.BAD_REQUEST,
+			});
+		}
```
kalinkaaaa14, line 33: Such checks should be performed on validation layer

apps/backend/src/db/migrations/20250726190816_add_reviews_table.ts
kalinkaaaa14: Same about poi_id and route_id

vldkorin: > Same about poi_id and route_id - they are references to corresponding tables

i am waiting for these tables to be merged

liza-veis (changes requested):
Looks good overall, a few comments 😎

apps/backend/src/modules/reviews/review.controller.ts
```diff
@@ -0,0 +1,154 @@
...
+	type ReviewRequestDto,
+	type ReviewResponseDto,
+} from "./libs/types/types.js";
+import { reviewCreateValidationSchema } from "./libs/validation-schemas/validation-schemas.js";
+
+/**
+ * @swagger
+ * components:
+ *   schemas:
+ *     ReviewRequestDto:
```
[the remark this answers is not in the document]
liza-veis: oh, actually let's leave it as it is

apps/backend/src/modules/reviews/review.service.ts
```diff
@@ -0,0 +1,40 @@
...
+		this.reviewRepository = reviewRepository;
+	}
+
+	public async create(payload: ReviewRequestDto): Promise<ReviewResponseDto> {
+		const reviewEntity = ReviewEntity.initializeNew({
+			content: payload.content,
+			likesCount: 0,
+			poiId: payload.poiId ?? null,
+			routeId: payload.routeId ?? null,
+			userId: payload.userId,
```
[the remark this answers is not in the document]
vldkorin: > We already have a user injected into the request, so we can access it in the controller and pass it there

So, i need to create something like ReviewCreatePayload, because i deleted userId from ReviewRequestDto?

packages\shared\src\modules\reviews\libs\types\review-create-payload.type.ts
```ts
type ReviewCreatePayload = {
	content: string;
	poiId: null | number;
	routeId: null | number;
	userId: number;
};

export { type ReviewCreatePayload };

```
And then
```ts
import { type CollectionResult, type Service } from "~/libs/types/types.js";
import { ReviewEntity } from "~/modules/reviews/review.entity.js";
import { type ReviewRepository } from "~/modules/reviews/review.repository.js";

import {
	type ReviewCreatePayload,
	type ReviewResponseDto,
} from "./libs/types/types.js";

class ReviewService implements Service {
	private reviewRepository: ReviewRepository;

	public constructor(reviewRepository: ReviewRepository) {
		this.reviewRepository = reviewRepository;
	}

	public async create(
		payload: ReviewCreatePayload,
	): Promise<ReviewResponseDto> {
		const reviewEntity = ReviewEntity.initializeNew({
			content: payload.content,
			likesCount: 0,
			poiId: payload.poiId ?? null,
			routeId: payload.routeId ?? null,
			userId: payload.userId,
		});

		const item = await this.reviewRepository.create(reviewEntity);

		return item.toObject();
	}
```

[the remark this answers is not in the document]
vldkorin: @liza-veis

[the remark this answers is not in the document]
vldkorin: @liza-veis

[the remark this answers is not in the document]
vldkorin: @kalinkaaaa14 @lavandosya 
What do you think?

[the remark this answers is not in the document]
liza-veis: Let's use casting for Authenticated User instead of non non-null assertion where we need this

PR #93, opened 2025-07-28
feat(frontend): add carousel component ss-87

### LINKED ISSUE #87: feat: Create common carousel component
### What feature?

- [x] Implement a common carousel component that scrolls to the right.
- [x] Add an image carousel somewhere on the main page for testing.

Notes: 
It should not be that hard to implement with pure html/css, so no libraries are needed for now.

### Screenshots

Image

vyacheslavKhoroshev:
- Spring bounce animations when reaching boundaries
- Slingshot effects on overdrag
- Smooth transitions between states
- Inertia effect
- Responsive layout with flexbox
- Automatic element sizing (5 per row)
- Hidden scrollbar for clean UI
- Snap functionality;

Closes #87


apps/frontend/src/libs/components/carousel/libs/enums/carausel-config.enum.ts
```diff
@@ -0,0 +1,18 @@
+const carouselConfig = {
```
liza-veis, line 1: should be in constants, not enum. Enums are not just objects; they are used to enumerate possible values for a specific constant, not to keep a set of constants and values
```suggestion
const CAROUSEL_CONFIG = {
```

apps/frontend/src/libs/components/carousel/libs/helpers/get-carousel-parameters.ts
```diff
@@ -0,0 +1,51 @@
+import { carouselConfig } from "../enums/enums.js";
+import { type CarouselParameters } from "../types/types.js";
+
+const initialCarouselParameters: CarouselParameters = {
```
liza-veis, line 4: Use uppercase for non-computed top level constants

apps/frontend/src/libs/components/carousel/libs/helpers/get-carousel-parameters.ts
liza-veis, line 1: file name for helpers should end with `.helper.ts`

liza-veis (changes requested):

lavandosya (approved):

PR #104, opened 2025-07-29
feat(backend/shared): add routes module ss-69

timosh-yura:
Added:
- Tables for routes in DB
- /routes controller
- CRUD operations for the routes module
- Data validation for the routes module
- Unit tests covering new functionality
image


vyacheslavKhoroshev:
> Please check comments about my PR. [#52](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/59) Because we have the same mistakes. For example: using HTTPMethodEnum, methods "Find", ZodType in validation schema. First of all you should pull changes for my branch because my PR almost merged.

Upd, already merged

apps/backend/src/modules/routes/routes.controller.test.ts
```diff
@@ -0,0 +1,162 @@
...
+	description: "Test route description",
+	id: FIRST_POI_ID,
+	name: "Test Route",
+	pois: [
+		{ id: FIRST_POI_ID, visitOrder: FIRST_VISIT_ORDER },
+		{ id: SECOND_POI_ID, visitOrder: SECOND_VISIT_ORDER },
+	],
+};
+
+const mockUser = { email: "test@example.com", id: 1 }; // Добавляем mock пользователя
```
h4vefundieyoung, line 32: really? 🤣

timosh-yura: 🥲

apps/backend/src/modules/routes/routes.controller.ts
```diff
@@ -0,0 +1,282 @@
...
+	 *             schema:
+	 *               type: object
+	 *               properties:
+	 *                 data:
+	 *                   type: array
+	 *                   items:
+	 *                     $ref: '#/components/schemas/Route'
+	 * */
+
+	public async findAll(): Promise<APIHandlerResponse<unknown[]>> {
```
my-derkach, line 208: My recommendation would be to try to avoid just `unknown[]`, because it is quite not obvious. Try maybe to replace it with `RoutesResponseDto[]`, especially when you have something like this

apps/backend/src/modules/routes/routes.controller.ts
```diff
@@ -0,0 +1,282 @@
...
+	 *         description: Route was not found
+	 */
+
+	public async patch(
+		options: APIHandlerOptions<{
+			body: RoutesRequestDto;
+			params: { id: string };
+		}>,
+	): Promise<APIHandlerResponse<RoutesResponseDto>> {
+		const id = Number(options.params.id);
```
my-derkach, line 266: I think that here can be the possible improve to make perfect validation of ID, because converting string to number can `NaN` if the id is not valid number as a string. It would be better to add additional checking of being `NaN` for id after converting string

timosh-yura:
Not sure about ID validation, I suggest using ZodType like its used for body
`validation: { body: routesCreateValidationSchema }`
↓
` validation: { body: routesCreateValidationSchema, params: idValidation}`
but it requires modifying BaseController's handler and thats out of the issue's scope
Implemented everything else, thanks for the review

apps/backend/src/modules/routes/routes.controller.ts
```diff
@@ -0,0 +1,282 @@
...
+	 *         schema:
+	 *           type: string
+	 *     responses:
+	 *       200:
+	 *         description: Route deleted successfully
+	 *       404:
+	 *         description: Route was not found
+	 */
+
+	public async delete(
```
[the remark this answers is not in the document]
liza-veis: This is not fixed yet

liza-veis (changes requested):

liza-veis (changes requested):

apps/backend/src/modules/routes/routes.repository.ts
```diff
@@ -0,0 +1,99 @@
...
+
+		const patchData = Object.fromEntries(
+			Object.entries({ description, name }).filter(
+				([, value]) => value !== undefined,
+			),
+		);
+
+		await this.routesModel.query().patchAndFetchById(id, patchData).execute();
+
+		return (await this.findById(id)) as RoutesResponseDto;
```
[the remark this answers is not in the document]
timosh-yura: patchAndFetchById doesnt include "pois" row, should I just not return it?

timosh-yura:
Updated

apps/backend/src/modules/routes/routes.service.ts
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		return route;
+	}
+
+	public async patch(
+		id: number,
+		payload: RoutesRequestPatchDto,
+	): Promise<RoutesResponseDto> {
+		await this.findById(id);
```
liza-veis, line 75: just check if updatedRoute is present and throw error if not, we need to eliminate unnecessary calls to db and we also need the same check for delete method - if false was returned, the not found error should be thrown
```suggestion
```

apps/backend/src/modules/routes/routes.repository.ts
```diff
@@ -0,0 +1,83 @@
...
+
+class RoutesRepository implements Repository {
+	private routesModel: typeof RoutesModel;
+
+	public constructor(routesModel: typeof RoutesModel) {
+		this.routesModel = routesModel;
+	}
+
+	public async create(entity: RouteEntity): Promise<RoutesResponseDto> {
+		const insertData = entity.toNewObject();
```
liza-veis, line 20: let's do it the same way as in other places, destructure needed fields and pass them explicitly to insertGraph

apps/backend/src/modules/routes/routes.repository.ts
```diff
@@ -0,0 +1,83 @@
...
+		const result = (await this.routesModel
+			.query()
+			.insertGraph(insertData, { relate: ["pois"] })
+			.returning([
+				"id",
+				"name",
+				"description",
+			])) as unknown as RoutesResponseDto;
+
+		return result;
```
liza-veis, line 31: Check out how it's done in other places. We need to use entity initialize and return the entity from the repository level whenever we return an object. Also, we need to get rid of the typecast here; the entity should provide the necessary type. And the same for all other methods in the repository.

 Let me know if there are other issues with types here, and typecast is needed. Also, don't forget to follow naming conventions 

```suggestion
		const route = await this.routesModel
			.query()
			.insertGraph(insertData, { relate: ["pois"] })
			.returning([
				"id",
				"name",
				"description",
			]));

		return RoutesEntity.initialize(route);
```

liza-veis (changes requested):
A few more comments, but I hope we will merge this tomorrow 🤞

apps/backend/src/modules/routes/routes.repository.ts
```diff
@@ -0,0 +1,83 @@
...
+
+		return (await this.findById(id)) as RoutesResponseDto;
+	}
+
+	private queryRoute(): QueryBuilder<RoutesModel, RoutesModel[]> {
+		return this.routesModel
+			.query()
+			.withGraphFetched("pois(selectIdOrder)")
+			.modifiers({
+				selectIdOrder(builder) {
```
liza-veis, line 75:
```suggestion
				selectPoiIdOrder(builder) {
```

apps/backend/src/modules/routes/routes.repository.ts
```diff
@@ -0,0 +1,83 @@
...
+	): Promise<RoutesResponseDto> {
+		await this.routesModel
+			.query()
+			.patchAndFetchById(id, entity as Partial<RoutesModel>)
+			.execute();
+
+		return (await this.findById(id)) as RoutesResponseDto;
+	}
+
+	private queryRoute(): QueryBuilder<RoutesModel, RoutesModel[]> {
```
liza-veis, line 70: Also, I see you have separate getById and queryRoute, let's use only one for now and avoid using queryById just to check existance. getById usually returns the whole information, and in case we need something with less data, we will create a method that will query less, or pass options to getById. But in our case, I beleive we can eliminate unnecessary query to check existance, so no additional method is needed for now and we only need getById which will return route with pois

liza-veis (changes requested):

apps/backend/src/modules/points-of-interest/points-of-interest.repository.ts
```diff
@@ -57,6 +62,11 @@ class PointsOfInterestRepository implements Repository {
 					"ST_AsGeoJSON(location)::json as location",
 				),
 			])
+			.modify((builder) => {
+				if (options && options.ids) {
```
liza-veis, line 66:
```suggestion
				if (options?.ids) {
```

apps/backend/src/modules/routes/routes.service.ts
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		return item.toObject();
+	}
+
+	private async ensurePoisExist(pois: number[]): Promise<void> {
+		const filteredPois = await this.pointsOfInterestService.findAll({
+			ids: pois,
+			latitude: 0,
+			longitude: 0,
```
liza-veis, line 104: this params are not required as i remember, we don't need to pass them here
```suggestion
```

packages/shared/src/modules/points-of-interest/libs/types/points-of-interest-search-query.type.ts
```diff
@@ -1,4 +1,5 @@
 type PointsOfInterestSearchQuery = {
+	ids?: number[];
```
liza-veis, line 2: they are a part of options, but not search query
```suggestion
```

apps/backend/src/modules/points-of-interest/points-of-interest.service.ts
```diff
@@ -50,13 +50,13 @@ class PointsOfInterestService implements Service {
 	}
 
 	public async findAll(
-		query: null | PointsOfInterestSearchQuery,
+		query: null | PointsOfInterestFindAllOptions,
```
liza-veis, line 53: let's rename it to options now
```suggestion
		options: null | PointsOfInterestFindAllOptions,
```

liza-veis (changes requested):

apps/backend/src/modules/routes/routes.service.ts
timosh-yura: They are required

apps/backend/src/modules/routes/routes.repository.ts
```diff
@@ -0,0 +1,89 @@
...
+				},
+			})
+			.select("routes.id", "routes.name", "routes.description");
+
+		if (!JSON.stringify(route)) {
+			return null;
+		}
+
+		return RoutesEntity.initialize(route);
+	}
```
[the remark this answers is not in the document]
timosh-yura: I`m not sure why, but eslint thinks its always truthy
 81:10  error  Unnecessary conditional, value is always truthy  @typescript-eslint/no-unnecessary-condition
Should I typecast route as RoutesModel | null then?

apps/backend/src/modules/routes/routes.service.ts
liza-veis: By logic, they are not required. I see the query interface requires either null or query params. I think since we are extending the interface, we can make them optional in the interface

apps/backend/src/modules/routes/routes.repository.ts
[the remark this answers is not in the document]
liza-veis: I see you have fixed it, so I think it's fine now

liza-veis:
LGTM, only this comment still needs to be resolved

https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/104#discussion_r2257853268

apps/backend/src/modules/routes/routes.service.ts
kalinkaaaa14: Passing them as 0 is invalid logic and in the future can lead to unexpected behavior. We need to pass latitude and longitude only if we want to find pois to exact geo coordinates. 

Let's make the coordinates as optional fields in the type, or add a new method is pointsOfInterestService findPoisByIds which will include the logic that you want - what's easier for you and requires minimal changes

apps/backend/src/modules/routes/routes.repository.ts
```diff
@@ -0,0 +1,89 @@
...
+					builder.select("points_of_interest.id", "routes_to_pois.visit_order");
+				},
+			})
+			.select(
+				"routes.id",
+				"routes.name",
+				"routes.description",
+			)) as null | RoutesModel;
+
+		return route ? RoutesEntity.initialize(route) : null;
```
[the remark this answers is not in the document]
timosh-yura: if I remove withGraphFetched it wont return pois in response body, is it okay?
with withGraphFetched it`s not itterable

apps/backend/src/modules/routes/routes.controller.ts
```diff
@@ -0,0 +1,299 @@
...
+			handler: this.patch.bind(this),
+			method: "PATCH",
+			path: "/:id",
+			validation: { body: routesUpdateValidationSchema },
+		});
+	}
+
+	/**
+	 * @swagger
+	 * /routes:
```
liza-veis, line 80: Add bearer authentication to every route here, otherwise it won't path token to the request and we won't be able to test it through swagger

```
 *     security:
 *       - bearerAuth: []
```

liza-veis (approved):

Issue #110, opened 2025-07-29
fix: Validation message shows "First name is required" for short name instead of indicating minimum length in sign up form


yuliiaaaaa:
### Describe your issue

When only 1 character is entered in the "First Name" field during sign up, the validation error says that the field is required, even though it is not empty. This is misleading for the user

### Steps to reproduce

1. Go to the page https://smart-scapes.com
2. Open the Sign up page
3. Enter 1 character in "First Name" field 
4. Observe the validation message

### What was the actual result?

The validation displays "First name is required" error message, when the field is not empty and contains 1 character

### What was the expected result?

The validation should display a clear error message such as: Name must be at least 2 characters long.

### Put here any screenshots or videos (optional)

<img width="521" height="161" alt="Image" src="https://github.com/user-attachments/assets/e6990907-26bc-4d37-b8bf-5f6a9ffa2827" />

### What OS are you seeing the problem on?

Windows

### What browsers are you seeing the problem on?

Chrome

### Severity

Trivial (S5)

PR #119, opened 2025-07-29
fix(shared): add validation messages when the input field is empty (ss-125) ss-110

my-derkach:
fix bug #110

### DIFF
--- a/packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
+++ b/packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
@@ -2,6 +2,7 @@ import { UserValidationRule } from "./user-validation-rule.enum.js";
 
 const UserValidationMessage = {
 	EMAIL_MAXIMUM_LENGTH: `Email must be at most ${String(UserValidationRule.MAX_LENGTH)} characters long.`,
+	EMAIL_MINIMUM_LENGTH: `Email must be at least ${String(UserValidationRule.EMAIL_MINIMUM_LENGTH)} characters long.`,
 	EMAIL_REQUIRED: "Email is required.",
 	EMAIL_WRONG: "Email format is wrong.",
 	FIRST_NAME_MAXIMUM_LENGTH: `First name must be at most ${String(UserValidationRule.MAX_LENGTH)} characters long`,
--- a/packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
+++ b/packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
@@ -17,13 +17,19 @@ const userSignUp = z
 				error: UserValidationMessage.PASSWORD_MAXIMUM_LENGTH,
 			}),
 		email: z
-			.email({
-				error: UserValidationMessage.EMAIL_WRONG,
-			})
+			.string()
 			.trim()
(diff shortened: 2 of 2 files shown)


my-derkach:
Yeah, nonempty is available on zod 3, but i will rewrite validations with min rn

my-derkach:
I have a question, Yulia shared with me all details about validation rules of sign up form, and it works fine for now, except email field:
```
email: z
			.string()
			.trim()
			.min(UserValidationRule.REQUIRED_STRING_MIN_LENGTH, {
				error: UserValidationMessage.EMAIL_REQUIRED,
			})
			.email({
				error: UserValidationMessage.EMAIL_WRONG,
			}),
```

I am trying to do something like this, so that if field is empty i got that the email is required and only after any character was entered i got wrong format, and it was working, but the problem occurred when I am trying to commit this changes:
image
From that point i need to remove string and put email as first, but with such approach we will not be able to get anything else except wrong email format even when the field is empty. What we are going to do?
Maybe we can try to use refine() for email with the custom regex?

vyacheslavKhoroshev:
> I have a question, Yulia shared with me all details about validation rules of sign up form, and it works fine for now, except email field:
> 
> ```
> email: z
> 			.string()
> 			.trim()
> 			.min(UserValidationRule.REQUIRED_STRING_MIN_LENGTH, {
> 				error: UserValidationMessage.EMAIL_REQUIRED,
> 			})
> 			.email({
> 				error: UserValidationMessage.EMAIL_WRONG,
> 			}),
> ```
> 
> I am trying to do something like this, so that if field is empty i got that the email is required and only after any character was entered i got wrong format, and it was working, but the problem occurred when I am trying to commit this changes: image From that point i need to remove string and put email as first, but with such approach we will not be able to get anything else except wrong email format even when the field is empty. What we are going to do? Maybe we can try to use refine() for email with the custom regex?

use z.email() instead z.string().email().

my-derkach:
Yeah, but if we will use such approach we will not get such message as "Email is required", because it firstly will check format firstly

> > I have a question, Yulia shared with me all details about validation rules of sign up form, and it works fine for now, except email field:
> > ```
> > email: z
> > 			.string()
> > 			.trim()
> > 			.min(UserValidationRule.REQUIRED_STRING_MIN_LENGTH, {
> > 				error: UserValidationMessage.EMAIL_REQUIRED,
> > 			})
> > 			.email({
> > 				error: UserValidationMessage.EMAIL_WRONG,
> > 			}),
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I am trying to do something like this, so that if field is empty i got that the email is required and only after any character was entered i got wrong format, and it was working, but the problem occurred when I am trying to commit this changes: image From that point i need to remove string and put email as first, but with such approach we will not be able to get anything else except wrong email format even when the field is empty. What we are going to do? Maybe we can try to use refine() for email with the custom regex?
> 
> use z.email() instead z.string().email().

Yeah, but with such implementation you will not get message "Email is required", because it will always check firstly the format

my-derkach:
Yeah, I am also thinking that there is no other way to do it except such approach with regex, just interested what coaches thinking about that: is it crucial to have this "Email is required" message

vyacheslavKhoroshev:
image

i found another approach, test it pls

liza-veis (approved):

PR #121, opened 2025-07-29
feat(backend/shared): add postgis ss-112

### LINKED ISSUE #112: feat: Postgis extension for Postgres
- [x] Install PostGIS extension on the database through the migration
- [x] Add required column location to poi table, which will be geometry type (exists after installing the extension)
- [x] Update the API of POI creation to save correct location for the poi. From the frontend - expected payload will be like this
```
{
  "name": "Central Park",
  "location": {
    "type": "Point",
    "coordinates": [-73.968285, 40.785091] 
  }
}
```
- [x] Update schema in readme

vyacheslavKhoroshev:
Closes #112

- Install PostGIS extension on the database through the migration

- Add required column location to poi table, which will be geometry type

- Update the API of POI creation to save correct location for the poi.

- Update schema in readme


apps/backend/src/modules/points-of-interest/points-of-interest.repository.ts
```diff
@@ -35,51 +44,95 @@ class PointsOfInterestRepository implements Repository {
...
+			locationString,
+		) as PointsOfInterestLocation;
+
+		return PointsOfInterestEntity.initialize({
+			createdAt: point.createdAt,
+			id: point.id,
+			location: parsedLocation,
+			name: point.name,
+			updatedAt: point.updatedAt,
+		});
```
liza-veis, line 135: I believe this method won't be extensible. We would need to make it too versatile and add conditions in it in the future to handle different return types, which we most probably will have. Let's handle this logic in each method separately. Also, about the condition where you check that point.location is of type string, i believe we can rely on a single response format from the db, correct me if that's not the case, so we don't want to have additional checks

liza-veis (changes requested):
Good work, left a few comments 🐈

packages/shared/src/modules/points-of-interest/libs/types/location.type.ts
liza-veis, line 1: Also, I believe this will be a generic type for the overall location of a type point, not only for points of interest, so I think it's a good idea to generalize here - e.g., PointLocation. In the future, if we add different types of locations, we will add types like StringLineLocation, and then if we need a general location type, we may use a union of them. It may look something like this with the oversight of the future (for now only PointLocation is needed)

```ts
type PointGeometry = {
  coordinates: [number, number];
  type: LocationType.POINT;
};

type LineStringGeometry = {
  coordinates: [number, number][];
  type: LocationType.LINE_STRING;
};

type LocationGeometry = PointGeometry | LineStringGeometry;
```

apps/backend/src/modules/points-of-interest/points-of-interest.repository.ts
vyacheslavKhoroshev: I dove deeper and found a more elegant solution)

lavandosya (approved):

PR #122, opened 2025-07-30
feat(backend): add user auth flow ss-26

h4vefundieyoung:
- /auth/authenticated-user endpoint has been implemented.
- swagger api has been added to the white routes list.
- added a few auth-flow test cases

https://github.com/user-attachments/assets/337ee00c-897a-40b9-817a-344146f6e7f0


apps/backend/src/libs/modules/plugins/auth/auth.plugin.ts
```diff
@@ -22,13 +23,29 @@ type WhiteRoute = {
...
+
+				const isRootPath = path.includes("/*");
+
+				if (!isRootPath) {
+					return url.endsWith(path);
+				}
+
+				const firstCharIndex = 0;
+				const lastCharIndex = -2;
+				const rootPath = `${basePath}${path.slice(firstCharIndex, lastCharIndex)}`;
```
liza-veis, line 45: Could you describe in more detaill, what this logic should handle

h4vefundieyoung: > Could you describe in more detaill, what this logic should handle

it gives an opportunity to add endpoints this way - "/xyz/*" so all the xyz-related endpoints will be handled the same way. the reason behind it - static resources, the swagger page itself fetches different resources, so we can't just add it to white list like a { path: "/documentation", method: "get" }, and adding every resource separately is just redundant as for me.

apps/backend/src/libs/modules/plugins/auth/auth.plugin.ts
```diff
@@ -22,13 +23,30 @@ type WhiteRoute = {
...
 		const { headers, method, url } = request;
 
 		const isWhiteListRoute = whiteRoutesList.some(
-			({ method: _method, path }) => url.endsWith(path) && _method === method,
+			({ method: _method, path }) => {
+				if (_method !== method) {
+					return false;
+				}
+
+				const isRootPath = path.includes("/*");
```
liza-veis, line 38: let's create a constant for "/*"

apps/backend/src/libs/modules/plugins/auth/auth.plugin.ts
```diff
@@ -22,13 +23,30 @@ type WhiteRoute = {
...
 const auth = (
 	app: FastifyInstance,
-	{ whiteRoutesList }: PluginOptions,
+
+	{ basePath, whiteRoutesList }: PluginOptions,
 ): void => {
 	const requestHandler = async (request: FastifyRequest): Promise<void> => {
 		const { headers, method, url } = request;
 
 		const isWhiteListRoute = whiteRoutesList.some(
```
liza-veis, line 32: let's move this logic to separate helper - checkIsWhiteRoute, which will accept white routes list and current route (method, path) to compare against

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -194,7 +194,11 @@ class BaseServerApplication implements ServerApplication {
 	}
 
 	private async initPlugins(): Promise<void> {
-		await this.app.register(authPlugin, { whiteRoutesList: WHITE_ROUTES });
+		const basePath = this.apis[0]?.basePath ?? "";
```
liza-veis, line 197: seems like we should consider multiple APIs, so let's iterate through them instead of picking the first one. We have something similar for docs generation here

liza-veis (changes requested):

apps/backend/src/libs/modules/plugins/auth/auth.plugin.ts
h4vefundieyoung: reasonable

apps/backend/src/libs/modules/server-application/libs/constants/constants.ts
```diff
@@ -2,15 +2,21 @@ import { APIPath, AuthApiPath } from "~/libs/enums/enums.js";
...
-const WHITE_ROUTES: WhiteRoute[] = [
-	{
-		method: "POST",
-		path: `${APIPath.AUTH}${AuthApiPath.SIGN_IN}`,
-	},
-	{
-		method: "POST",
-		path: `${APIPath.AUTH}${AuthApiPath.SIGN_UP}`,
-	},
+const WHITE_ROUTES: WhiteRoute[][] = [
```
liza-veis, line 5: We can't be sure in index match for apis. If we want to keep them isolated for the api we need to move them to the api class itself. To store it there as a property

liza-veis (changes requested):

apps/backend/src/libs/modules/server-application/base-server-application-api.ts
```diff
@@ -70,6 +75,15 @@ class BaseServerApplicationApi implements ServerApplicationApi {
...
 	}
+
+	public injectWhiteRoutes(whiteRoutes: WhiteRoute[]): void {
+		const injectedRoutes = whiteRoutes.map((whiteRoute) => ({
+			...whiteRoute,
+			path: `${this.basePath}${whiteRoute.path}`,
+		}));
+
+		this.whiteRoutes = [...this.whiteRoutes, ...injectedRoutes];
+	}
```
liza-veis, line 86: For the white routes, it's more logical if they were defined by the api itself, not injected from the outside since it's not an external dependency or service, but rather needed internal api setup. So what I meant is that we basically need the same flow as with the documentation here. We define the config and setup right here and keep white routes here without even constant for now, since the config is not that large for now.

liza-veis (changes requested):

apps/backend/src/libs/modules/server-application/base-server-application-api.ts
h4vefundieyoung: @liza-veis But that way all api instances will share basically the same routing config, the only difference is basePath. Am I getting it right?

liza-veis: No, what I was trying to point outis that we want our apis to define white routes separately. Even if the config is repetitive, since there is difference in api routes in different apis, they should define white lists for themselves

h4vefundieyoung: I see no other solutions except DI, whether it'll be constructor or injector method.

liza-veis (approved):
I meant it slightly differently, to have those paths defined in the api constructor itself, without any outside control, but this approach looks good to me to 🐈

liza-veis (approved):

PR #131, opened 2025-07-30
feat(shared/backend): add notifications module ss-73

my-derkach:
On the backend side:
* add new table notifications to db
* create notification module which consists of:
     * types for request and response dto, validation schemas for creating notification (from shared)
     * entity, model, repository, service, controller
     * it was covered with tests: entity, repository, service, controller
     * add swagger documentation
     
In shared:
* create notification module where define:
     * enums: validation rules and messages, define entity and notification types, notification api path
     * types: request and response dto
     * validation-schems: notification create validation schema
     
Root:
* make mermaid diagram of notifications

### DIFF
--- a/apps/backend/src/db/migrations/20250728080941_add_notifications_table.ts
+++ b/apps/backend/src/db/migrations/20250728080941_add_notifications_table.ts
@@ -54,7 +54,6 @@ async function up(knex: Knex): Promise<void> {
 
 			table
 				.integer(ColumnName.USER_ID)
-				.unsigned()
 				.notNullable()
 				.references(ColumnName.ID)
 				.inTable(USERS_TABLE_NAME)


packages/shared/src/modules/notifications/libs/validation-schemas/notification-create.validation-schema.ts
```diff
@@ -0,0 +1,47 @@
...
+		}),
+	entityId: z
+		.number()
+		.int({
+			message: NotificationValidationMessage.ENTITY_ID_MUST_BE_NUMBER,
+		})
+		.positive({
+			message: NotificationValidationMessage.ENTITY_ID_MUST_BE_POSITIVE,
+		}),
+	entityType: z.enum(Object.values(EntityType) as [string, ...string[]], {
```
liza-veis, line 28: what do we need this typecast for

packages/shared/src/modules/notifications/libs/enums/notification-api-path.enum.ts
```diff
@@ -0,0 +1,6 @@
+const NotificationApiPath = {
+	$ID: ":/id",
```
liza-veis, line 2: shouldn't it be
```suggestion
	$ID: "/:id",
```

packages/shared/src/modules/notifications/libs/enums/entity-type.enum.ts
```diff
@@ -0,0 +1,5 @@
+const EntityType = {
```
liza-veis, line 1:
```suggestion
const NotificationEntityType = {
```
don't forget about file name

apps/backend/src/modules/notifications/notification.model.ts
```diff
@@ -0,0 +1,19 @@
...
+
+class NotificationModel extends AbstractModel {
+	public static override get tableName(): string {
+		return DatabaseTableName.NOTIFICATIONS;
+	}
+
+	public content!: string;
+	public entityId!: number;
+	public entityType!: string;
+	public notificationType!: string;
```
liza-veis, line 14: let's use valueof enums here

apps/backend/src/modules/notifications/libs/enums/notification-column-name.enum.ts
```diff
@@ -0,0 +1,18 @@
+const NotificationColumnName = {
```
liza-veis, line 1: What do we need this for, i think this can be removed

apps/backend/src/db/migrations/20250728080941_add_notifications_table.ts
```diff
@@ -0,0 +1,71 @@
...
+	});
+}
+
+async function up(knex: Knex): Promise<void> {
+	await knex.raw(`
+		CREATE TYPE notification_type AS ENUM ('${Object.values(NotificationType).join("', '")}')
+	`);
+
+	await knex.raw(`
+		CREATE TYPE entity_type AS ENUM ('${Object.values(EntityType).join("', '")}')
```
liza-veis, line 39:
```suggestion
		CREATE TYPE notification_entity_type AS ENUM ('${Object.values(EntityType).join("', '")}')
```

apps/backend/src/db/migrations/20250728080941_add_notifications_table.ts
```diff
@@ -0,0 +1,71 @@
...
+	READ_AT: "read_at",
+	UPDATED_AT: "updated_at",
+	USER_ID: "user_id",
+} as const;
+
+const NotificationType = {
+	USER_FOLLOWED: "user_followed",
+} as const;
+
+const EntityType = {
```
liza-veis, line 22:
```suggestion
const NotificationEntityType = {
```

liza-veis (changes requested):
A few comments, but overall great job 😎

apps/backend/src/db/migrations/20250728080941_add_notifications_table.ts
```diff
@@ -0,0 +1,71 @@
...
+			.notNullable()
+			.defaultTo(knex.fn.now());
+		table
+			.dateTime(ColumnName.UPDATED_AT)
+			.notNullable()
+			.defaultTo(knex.fn.now());
+
+		table
+			.integer(ColumnName.USER_ID)
+			.unsigned()
```
kalinkaaaa14, line 55: Not sure why it's needed to use .unsigned ?

kalinkaaaa14 (commented):
Good job 👍

packages/shared/src/modules/notifications/libs/validation-schemas/notification-create.validation-schema.ts
my-derkach: Initially I was thinking to use nativeEnum, but I do not define it as enum, so I decided to use z.enum(). As I understand the z.enum() requires a tuple of strings, so using such typecast to satisfy typing requirements

packages/shared/src/modules/notifications/libs/enums/entity-type.enum.ts
my-derkach: So, generally i just need to change type from EntityType to NotificationEntityType

liza-veis (approved):

apps/backend/src/db/migrations/20250728080941_add_notifications_table.ts
kalinkaaaa14: seems on Postgresql unsigned is ignored

kalinkaaaa14: So remove it please

my-derkach: Done)

PR #136, opened 2025-07-30
feat(backend/shared): add followers module ss-72

D1monB:
https://www.loom.com/share/dcaf06bf009d4b95aa1fb4c0c25c3282 - testing followers module
in the video the user with id1 has 'Bearer JWT' in the headers

- I created new user_follows table
- I expanded the swagger documentation
- Added new routes, and implemented follow and unfollow methods in controller, service and repository
- I added two many-to-many relations in the UserModel 'one for users who follow this user (followers), and one for users whom this user is following (following)'
- I used a join table user_follows to connect users with each other

I know I should have written or updated tests for all the changes, and I'm really sorry, I just don’t know how to write good tests yet


D1monB:
> Looks really good to me)

Thank you, at weekends i will try to write tests, i don't why, but, tests are the most difficult subject for me

apps/backend/src/modules/users/user.service.ts
```diff
@@ -77,6 +77,73 @@ class UserService implements Service {
 	): Promise<null | UserPasswordDetails> {
 		return await this.userRepository.findPasswordDetails(email);
 	}
+
+	public async follow(followerId: number, followingId: number): Promise<void> {
+		if (followerId === followingId) {
+			throw new UserError({
+				message: "Cannot follow yourself",
```
liza-veis, line 84: Use user exception message for all messages here

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -36,6 +37,18 @@ class UserController extends BaseController {
...
+			handler: this.follow.bind(this),
+			method: "POST",
+			path: UsersApiPath.FOLLOW,
+		});
+
+		this.addRoute({
+			handler: this.unfollow.bind(this),
+			method: "DELETE",
+			path: UsersApiPath.UNFOLLOW,
+		});
```
liza-veis, line 51: Instead of having this logic in the user module, let's create a separate one e.g., user-follows, and move everything there

apps/backend/src/db/migrations/20250730123508_add_user_follows_table.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+function up(knex: Knex): Promise<void> {
+	return knex.schema.createTable(TABLE_NAME, (table) => {
+		table.increments(ColumnName.ID).primary();
+
+		table
+			.integer(ColumnName.FOLLOWER_ID)
+			.notNullable()
+			.unsigned()
```
kalinkaaaa14, line 25: seems like .unsigned() is ignored in PostgreSQL

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -58,6 +58,39 @@ class UserRepository implements Repository {
...
+		followerId: number,
+		followingId: number,
+	): Promise<void> {
+		await this.userModel
+			.relatedQuery("following")
+			.for(followerId)
+			.relate(followingId);
+	}
+
+	public async isFollowing(
```
kalinkaaaa14, line 72:
```suggestion
	public async checkIsUserFollowing(
```

apps/backend/src/db/migrations/20250730123508_add_user_follows_table.ts
kalinkaaaa14: In the documentation of Postgres I see that the unsigned integer types are not supported 🤔 so pretty weird

kalinkaaaa14: but despite even this fact, follower_id is the references to id in users table, which is always positive value and even without adding this .unsigned - postgres won't allow you to store the negative reference to id which doesn't exist. So in my opinion adding this is redundant

apps/backend/src/libs/modules/server-application/server-application.ts
```diff
@@ -35,6 +36,8 @@ const apiV1 = new BaseServerApplicationApi(
 	...routeCategoryController.routes,
 	...routesController.routes,
 	...userController.routes,
+	...pointsOfInterestController.routes,
```
liza-veis, line 39: we already have it above
```suggestion
```

apps/backend/src/modules/user-follows/user-follows.controller.ts
```diff
@@ -0,0 +1,152 @@
...
+			method: "DELETE",
+			path: UserFollowsApiPath.UNFOLLOW,
+		});
+	}
+
+	/**
+	 * @swagger
+	 * /users/{userId}/follow:
+	 *   post:
+	 *     tags:
```
liza-veis, line 43: let's add bearerAuth to both swagger schemas
```
	 *     security:
	 *       - bearerAuth: []
```

apps/backend/src/modules/user-follows/user-follows.controller.ts
```diff
@@ -0,0 +1,152 @@
...
+	): Promise<APIHandlerResponse> {
+		const followingId = Number(request.body.followingId);
+		const followerId = Number(request.params.userId);
+		const userId = Number(request.user?.id);
+
+		if (userId !== followerId) {
+			return {
+				payload: null,
+				status: HTTPCode.UNAUTHORIZED,
+			};
```
liza-veis, line 87: let's throw UserFollowsError instead with status 401 and appropriate exception message as in other places, it will be handled and sent as approriate response by common error handler in server application

apps/backend/src/modules/user-follows/user-follows.controller.ts
```diff
@@ -0,0 +1,152 @@
...
+		request: APIHandlerOptions<
+			{
+				body: UserFollowsFollowRequestDto;
+				params: UserFollowsFollowParametersDto;
+			} & { user: UserGetByIdItemResponseDto }
+		>,
+	): Promise<APIHandlerResponse> {
+		const followingId = Number(request.body.followingId);
+		const followerId = Number(request.params.userId);
+		const userId = Number(request.user?.id);
```
liza-veis, line 81: We already know we will have user here, so let's cast it. As for the interface, we don't need to add user there, it's already included in type  and the same for unfollow
```suggestion
	public async follow(
		options: APIHandlerOptions<
			{
				body: UserFollowsFollowRequestDto;
				params: UserFollowsFollowParamsDto
			}
		>,
	): Promise<APIHandlerResponse> {
	    const { body, params, user } = options;
	    
		const followingId = Number(body.followingId);
		const followerId = Number(params.userId);
		const userId = (user as UserAuthResponseDto).id
```

apps/backend/src/modules/user-follows/user-follows.repository.ts
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	public async followUser(
+		followerId: number,
+		followingId: number,
+	): Promise<void> {
+		await UserFollowsModel.query().insert({
+			followerId,
+			followingId,
+		} as Partial<UserFollowsModel>);
```
liza-veis, line 28: do we need to cast it at all? I see we don't use casting in other places
```suggestion
		});
```

apps/backend/src/modules/user-follows/user-follows.repository.ts
```diff
@@ -0,0 +1,44 @@
...
+	public async unfollowUser(
+		followerId: number,
+		followingId: number,
+	): Promise<number> {
+		const deleteCount = await UserFollowsModel.query().delete().where({
+			followerId,
+			followingId,
+		});
+
+		return deleteCount;
```
liza-veis, line 40: let's return boolean as in other places
```suggestion
		return Boolean(deletedCount);
```

packages/shared/src/modules/user-follows/libs/enums/user-follows-api-path.enum.ts
```diff
@@ -0,0 +1,6 @@
+const UserFollowsApiPath = {
+	FOLLOW: "/:userId/followers",
+	UNFOLLOW: "/:userId/followers/:id",
```
liza-veis, line 3: for the api paths, we will name them not by meaning, but by actual path, here is the pattern
```suggestion
	$USER_ID_FOLLOWERS: "/:userId/followers",
	$USER_ID_FOLLOWERS_$ID: "/:userId/followers/:id",
```

packages/shared/src/modules/user-follows/libs/types/user-follows-follow-parameters-dto.ts
liza-veis, line 1: file name should end with .type.ts and the same for user-follows-follow-request-dto.ts and user-follows-unfollow-parameters-dto.ts

apps/backend/src/libs/modules/server-application/server-application.ts
D1monB: Sorry it was conflict, and just kept this controller

apps/backend/src/modules/user-follows/user-follows.repository.ts
D1monB: TS2589: Type instantiation is excessively deep and possibly infinite.

i added this because i got an error

liza-veis (changes requested):

packages/shared/src/modules/user-follows/libs/types/types.ts
```diff
@@ -0,0 +1,3 @@
+export { type UserFollowsFollowParametersDto } from "./user-follows-follow-parameters-dto.types.js";
+export { type UserFollowsFollowRequestDto } from "./user-follows-follow-request-dto.types.js";
+export { type UserFollowsUnfollowParametersDto } from "./user-follows-unfollow-parameters-dto.types.js";
```
liza-veis, line 3: Other than this, LGTM
```suggestion
export { type UserFollowsFollowParametersDto } from "./user-follows-follow-parameters-dto.type.js";
export { type UserFollowsFollowRequestDto } from "./user-follows-follow-request-dto.type.js";
export { type UserFollowsUnfollowParametersDto } from "./user-follows-unfollow-parameters-dto.type.js";
```

D1monB: Refactored

liza-veis (approved):
LGTM, approved 🐈

apps/backend/src/modules/user-follows/user-follows.repository.test.ts
```diff
@@ -0,0 +1,90 @@
...
+				{ followerId: TEST_FOLLOWER_ID, followingId: TEST_FOLLOWING_ID },
+			]);
+
+		await userFollowsRepository.followUser(TEST_FOLLOWER_ID, TEST_FOLLOWING_ID);
+	});
+
+	it("unfollowUser should delete follow relation and return true", async () => {
+		databaseTracker.on.delete("user_follows").response(DELETED_COUNT_ONE);
+
+		const deletedCount = Boolean(
```
kalinkaaaa14, line 70:
```suggestion
		const isDeleted = Boolean(
```

apps/backend/src/modules/user-follows/user-follows.repository.test.ts
```diff
@@ -0,0 +1,90 @@
...
+				TEST_FOLLOWING_ID,
+			),
+		);
+		assert.strictEqual(deletedCount, true);
+	});
+
+	it("unfollowUser should return false if no rows deleted", async () => {
+		databaseTracker.on.delete("user_follows").response(DELETED_COUNT_NONE);
+
+		const deletedCount = Boolean(
```
kalinkaaaa14, line 82:
```suggestion
		const isDeleted = Boolean(
```

apps/backend/src/modules/user-follows/user-follows.repository.ts
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	public async checkIsUserFollowing(
+		followerId: number,
+		followingId: number,
+	): Promise<boolean> {
+		const relation = await this.userFollowsModel
+			.query()
+			.where({ followerId, followingId })
+			.first();
```
kalinkaaaa14, line 17:
```suggestion
		.query()
  .select(1)
  .where({ followerId, followingId })
  .first();
```

packages/shared/src/modules/user-follows/libs/types/user-follows-follow-parameters-dto.type.ts
```diff
@@ -0,0 +1,5 @@
+type UserFollowsFollowParametersDto = {
```
kalinkaaaa14, line 1: + do not forget to change file name
```suggestion
type UserFollowsParamsDto = {
```

packages/shared/src/modules/user-follows/libs/types/user-follows-follow-request-dto.type.ts
```diff
@@ -0,0 +1,5 @@
+type UserFollowsFollowRequestDto = {
```
kalinkaaaa14, line 1: + change filename 

```suggestion
type UserFollowsRequestDto = {
```

packages/shared/src/modules/user-follows/libs/types/user-follows-unfollow-parameters-dto.type.ts
```diff
@@ -0,0 +1,6 @@
+type UserFollowsUnfollowParametersDto = {
```
kalinkaaaa14, line 1: + change filename

```suggestion
type UserUnfollowsParamsDto = {
```

kalinkaaaa14 (commented):
Left few small comments, other than that good job! 👍

packages/shared/src/modules/user-follows/libs/types/user-follows-follow-parameters-dto.type.ts
D1monB: I can keep UserFollowsParametersDto if needed

kalinkaaaa14: @D1monB yes, let's move on with UserFollowsParametersDto 👍

D1monB:
I hope I didn’t miss anything 😅

PR #137, opened 2025-07-30
fix(shared): add Latin letters validation for user names ss-129

### LINKED ISSUE #129: fix: Registration with numbers and no latin characters in fields "First name" and "Last name" is successful
### Describe your issue

The Sign Up form allows registration when the First name and Last name fields are filled with numbers or with characters that are not Latin letters

### Steps to reproduce

1. Go to https://smart-scapes.com/
2. Open the Sign Up page
3. Enter numbers (e.g., 123) or non latin characters into First name and Last name fields
4. Fill in the other required fields with valid data
5. Submit the form

Registration is completed successfully

Validation should prevent using only numbers or no latin characters in the First name and Last name fields

Image

Windows

Chrome

### Severity

Major (S3)

vyacheslavKhoroshev:
fix(shared): add Latin letters validation for user names

- Add isLatinLetters helper function to validate names
- Block Cyrillic characters and numbers in firstName/lastName fields
- Update validation schema with proper error messages

Closes #129


packages/shared/src/libs/helpers/is-latin-letters/is-latin-letters.helper.ts
```diff
@@ -0,0 +1,7 @@
+const isLatinLetters = (value: string): boolean => {
```
kalinkaaaa14, line 1:
```suggestion
const checkIsLatinLetter = (value: string): boolean => {
```

PR #145, opened 2025-07-31
feat(shared): implement additional checking for email field with banned domain ss-144

my-derkach:
Implemented additional check for email validation schema to disallow a user to register with banned domains

### DIFF
--- a/packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
+++ b/packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
@@ -1,6 +1,7 @@
 import { UserValidationRule } from "./user-validation-rule.enum.js";
 
 const UserValidationMessage = {
+	EMAIL_DOMAIN_NOT_ALLOWED: "Email adresses with such domains are not allowed.",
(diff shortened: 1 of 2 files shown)


packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
```diff
@@ -1,6 +1,7 @@
 import { UserValidationRule } from "./user-validation-rule.enum.js";
 
 const UserValidationMessage = {
+	EMAIL_DOMAIN_NOT_ALLOWED: "Email adresses with such domains are not allowed.",
```
kalinkaaaa14, line 4:
```suggestion
	EMAIL_DOMAIN_NOT_ALLOWED: "Email addresses with such domains are not allowed.",
```

packages/shared/src/modules/users/libs/enums/user-validation-rule.enum.ts
```diff
@@ -1,4 +1,5 @@
 const UserValidationRule = {
+	BANNED_EMAIL_DOAMINS: [".ru"],
```
kalinkaaaa14, line 2:
```suggestion
	BANNED_EMAIL_DOMAINS: [".ru"],
```

PR #160, opened 2025-07-31
feat(backend): added group id for newly created users ss-113

DenisDemirov:
Added:

- Group entity
- Group repository
- Associated types and enums
- Updated Sign-In responce
- Added migration

Screenshot 2025-08-05 181009
Screenshot 2025-08-05 181029


DenisDemirov:
After automatic merge i faced with strange problem that blocked lint check 
зображення
I found than in auth-api (in frontend) signIn method used SignUpDto instead of SignInDto. I changed it, maybe its automatic merge artifact

apps/backend/src/modules/users/user.service.ts
```diff
@@ -26,17 +36,33 @@ class UserService implements Service {
...
+				status: HTTPCode.CONFLICT,
+			});
+		}
+
+		if (group.toObject().id === null) {
+			throw new UserError({
+				message: GroupExceptionMessage.GROUP_NOT_FOUND as string,
+				status: HTTPCode.CONFLICT,
+			});
+		}
```
lavandosya, line 59:
```suggestion
	if (group?.toObject().id == null) {
			throw new UserError({
				message: GroupExceptionMessage.GROUP_NOT_FOUND as string,
				status: HTTPCode.CONFLICT,
			});
		}
```

lavandosya:
> I found than in auth-api (in frontend) signIn method used SignUpDto instead of SignInDto. I changed it, maybe its automatic merge artifact

You can install the [GitLens](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens) and navigate through history to see when and by whom the change was introduced
image

apps/backend/src/db/migrations/20250730112137_user_group_id_required.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+async function up(knex: Knex): Promise<void> {
+	await knex(TABLE_NAME)
+		.whereNull(ColumnName.GROUP_ID)
+		.update({ [ColumnName.GROUP_ID]: 2 });
+
+	await knex.schema.alterTable(TABLE_NAME, (table) => {
+		table.integer(ColumnName.GROUP_ID).notNullable().alter();
+	});
```
liza-veis, line 22: Wrap multiple knex calls in a transaction

apps/backend/src/modules/groups/group.repository.ts
```diff
@@ -0,0 +1,43 @@
+import { type Repository } from "~/libs/types/types.js";
```
liza-veis, line 1: We are not planning any interface for groups on ui, and for this specific case, we want to join a group in a single query with the user, so I believe we won't need the groups repository and service. Nevertheless, I want to point out that it was done well 🐈

liza-veis (changes requested):
Not bad as for the backend task ✨ , left a few comments

kalinkaaaa14:
Also one more moment, it's not said in the ticket but we need to add it - the user should be returned not only with group, but also with permissions that are available to him. And the request for /authenticated-user should return permissions as well

apps/backend/src/modules/users/libs/types/user-password-details.type.ts
```diff
@@ -1,5 +1,12 @@
 type UserPasswordDetails = {
+	group: {
+		id: number;
+		key: string;
+		name?: string;
```
liza-veis, line 5: name shouldn't be optional when group is present
```suggestion
		name: string;
```

apps/backend/src/modules/groups/group.service.ts
```diff
@@ -0,0 +1,50 @@
...
+
+		return {
+			items: groups.map((group) => group.toObject()),
+		};
+	}
+
+	public async findByKey(key: string): Promise<GroupItemResponseDto | null> {
+		const group = await this.groupRepository.findByKey(key);
+
+		return group?.toObject() || null;
```
liza-veis, line 46: let's prefer nullish coalescing for such cases
```suggestion
		return group?.toObject() ?? null;
```

packages/shared/src/modules/users/libs/types/user-auth-response-dto.type.ts
```diff
@@ -1,6 +1,12 @@
 type UserAuthResponseDto = {
 	email: string;
 	firstName: string;
+	group?: {
+		id: number;
+		key: string;
+		name?: string;
+	};
```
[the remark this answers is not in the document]
liza-veis: Actually, I thought a bit, and now since group ID is required, we can't have a case where it's null, so let's make it required everywhere

DenisDemirov:
I know that its bad, but i despaired to fix it. I spend hours trying to fix test files and i faced with problem to check new data in user.repository.test.ts
зображення

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -29,34 +44,133 @@ class UserRepository implements Repository {
...
+	): Promise<null | UserGetByIdItemResponseDto> {
+		const user = await this.userModel
+			.query()
+			.where("email", email)
+			.withGraphJoined("group.permissions")
+			.first();
+
+		if (!user || !user.group || !user.group.permissions) {
+			return null;
+		}
```
liza-veis, line 88: The group will always be present and permissions should be set to empty array if not present
```suggestion
		if (!user) {
			return null;
		}
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -29,34 +44,133 @@ class UserRepository implements Repository {
...
+			},
+			groupId: user.groupId,
+			id: user.id,
+			lastName: user.lastName,
+			passwordHash: user.passwordHash,
+			passwordSalt: user.passwordSalt,
+		};
+	}
+
+	private mapUserModelToDto(
```
liza-veis, line 150: let's remove this method and do it separately for each method, also we need to return entity from the repository, not plain object
```suggestion
```

liza-veis (changes requested):

apps/backend/src/libs/modules/plugins/auth/auth.plugin.ts
```diff
@@ -44,7 +47,7 @@ const auth = (app: FastifyInstance, { whiteRoutes }: PluginOptions): void => {
 			throw new AuthError();
 		}
 
-		request.user = user;
+		request.user = user.toObject() as UserAuthResponseDto;
```
liza-veis, line 50: it should be a plain user here already
```suggestion
		request.user = user;
```

apps/backend/src/modules/groups/libs/helpers/to-group-entity.ts
```diff
@@ -0,0 +1,20 @@
...
+		name: model.name,
+		permissions: (model.permissions ?? []).map((permission) =>
+			PermissionEntity.initialize({
+				id: permission.id,
+				key: permission.key,
+				name: permission.name,
+			}),
+		),
+	});
+};
```
liza-veis, line 18: We don't need to call GroupEntity and PermissionEntity inside UserEntity, it's enough to just remap them as plain properties inside the UserEntity, remove this helper

apps/backend/src/modules/groups/group.service.ts
```diff
@@ -0,0 +1,50 @@
...
+		const item = await this.groupRepository.create(payload as GroupEntity);
+
+		return item.toObject();
+	}
+
+	public async find(id: number): Promise<GroupItemResponseDto> {
+		const group = await this.groupRepository.find(id);
+
+		if (!group) {
+			throw new Error(GroupExceptionMessage.GROUP_NOT_FOUND);
```
liza-veis, line 29: we need to create a custom exception
```suggestion
			throw new GroupError(GroupExceptionMessage.GROUP_NOT_FOUND);
```

apps/backend/src/modules/groups/group.service.ts
```diff
@@ -0,0 +1,50 @@
...
+
+class GroupService implements Service<GroupItemResponseDto> {
+	private groupRepository: GroupRepository;
+
+	public constructor(groupRepository: GroupRepository) {
+		this.groupRepository = groupRepository;
+	}
+
+	public async create(payload: unknown): Promise<GroupItemResponseDto> {
+		const item = await this.groupRepository.create(payload as GroupEntity);
```
liza-veis, line 20:
```suggestion
	public async create(payload: GroupEntity): Promise<GroupItemResponseDto> {
		const item = await this.groupRepository.create(payload);
```

apps/backend/src/modules/users/user.entity.ts
```diff
@@ -103,12 +123,25 @@ class UserEntity implements Entity {
...
+			id: number;
+			key: string;
+			name: string;
+			permissions: {
+				id: number;
+				key: string;
+				name: string;
+			}[];
+		};
+		groupId: null | number;
```
liza-veis, line 136: groupId can't be null, as well as the group itself; go through all places where you mention group and remove unnecessary null checks if any
```suggestion
		groupId: number;
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -1,62 +1,183 @@
...
 
 	public async create(entity: UserEntity): Promise<UserEntity> {
 		const { email, firstName, lastName, passwordHash, passwordSalt } =
 			entity.toNewObject();
 
+		const group = await this.groupService.findByKey(GroupKey.USERS);
+
+		if (!group) {
+			throw new Error(GroupExceptionMessage.GROUP_NOT_FOUND);
+		}
```
liza-veis, line 29: let's throw with error inside groupService findByKey

apps/backend/src/modules/users/user.service.ts
```diff
@@ -43,33 +48,45 @@ class UserService implements Service {
...
+		const newUser = await this.findById(item.toObject().id);
+
+		if (!newUser) {
+			throw new UserError({
+				message: CommonExceptionMessage.COMMON_EXCEPTION_MESSAGE,
+				status: HTTPCode.NOT_FOUND,
+			});
+		}
+
+		return newUser.toObject() as UserAuthResponseDto;
```
liza-veis, line 67: not sure what is this for, create should return the needed interface
```suggestion
		return item.toObject()
```

apps/backend/src/modules/users/user.service.ts
```diff
@@ -43,33 +48,45 @@ class UserService implements Service {
...
 
-	public async findById(
-		id: number,
-	): Promise<null | UserGetByIdItemResponseDto> {
+	public async findById(id: number): Promise<null | UserEntity> {
 		const user = await this.userRepository.findById(id);
 
-		return user ? user.toObject() : null;
+		return user;
 	}
```
liza-veis, line 90: this part should be returned back - we need to return entity from the repository and call entity to object from here to return plain object from the service

packages/shared/src/modules/groups/types/group-item-response-dto.type.ts
```diff
@@ -0,0 +1,6 @@
+type GroupItemResponseDto = {
+	id: number;
+	key: string;
```
liza-veis, line 3:
```suggestion
type GroupResponseDto = {
	id: number;
	key: ValueOf<typof GroupKey>;
```

liza-veis (changes requested):

DenisDemirov:
@liza-veis i think i am finished (for now)

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -3,60 +3,175 @@ import { type UserPasswordDetails } from "~/modules/users/libs/types/types.js";
...
+			group: {
+				id: user.group.id,
+				key: user.group.key,
+				name: user.group.name,
+				permissions: user.group.permissions.map((permission) => ({
+					id: permission.id,
+					key: permission.key,
+					name: permission.name,
+				})),
+			},
```
liza-veis, line 168: is there an issue with not using it the same way as in other methods? i'm ok with remapping overall, but just want it to be consistent
```suggestion
			group: GroupEntity.initialize(user.group).toObject()
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -3,60 +3,175 @@ import { type UserPasswordDetails } from "~/modules/users/libs/types/types.js";
...
+				"users.password_hash as passwordHash",
+				"users.password_salt as passwordSalt",
+				"users.group_id as groupId",
+			)
+			.withGraphJoined("group.permissions")
+			.where("users.email", email)
 			.first();
 
-		return user ?? null;
+		if (!user || !user.group || !user.group.permissions) {
```
liza-veis, line 153: we only need to check if user is present, the group is always present if you joined it and permissions should be an array by default. If it's possible for them to be null, it should be handled in GroupEntity.initialize 
```suggestion
		if (!user) {
```

liza-veis (changes requested):

apps/backend/src/modules/users/user.entity.ts
liza-veis: I forgot about this, in this case let's leave | null

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -3,60 +3,200 @@ import { type UserPasswordDetails } from "~/modules/users/libs/types/types.js";
...
 
-		return user ? UserEntity.initialize(user) : null;
+		if (!user) {
+			return null;
+		}
+
+		return UserEntity.initialize({
+			email: user.email,
+			firstName: user.firstName,
+			group: user.group
```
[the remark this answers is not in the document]
DenisDemirov: No, we do not have a group during user creation. We can get user with group from database only after creating a new user with id of group

apps/backend/src/modules/users/user.model.ts
```diff
@@ -1,21 +1,43 @@
...
 	}
+
 	public email!: string;
 
 	public firstName!: string;
 
+	public group?: GroupModel;
+
+	public groupId!: number;
+
```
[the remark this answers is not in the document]
DenisDemirov: Yes, but without group field inside user model i got a lot of errors because now ts is not expecting group inside UserModel in ansew 
зображення
Screenshot 2025-08-08 125422

packages/shared/src/modules/groups/types/group-item-dto.type.ts
```diff
@@ -0,0 +1,10 @@
+import { type PermissionItemDto } from "../../permission/types/types.js";
+
+type GroupItemDto = {
```
kalinkaaaa14, line 3: + do not forget about file name

```suggestion
type GroupItemWithPermissionsDto = GroupResponseDto & { permissions: PermissionItemDto[]; }
```

apps/backend/src/modules/users/libs/types/user-password-details.type.ts
```diff
@@ -1,5 +1,16 @@
...
+	group: {
+		id: number;
+		key: string;
+		name: string;
+		permissions: {
+			id: number;
+			key: string;
+			name: string;
+		}[];
+	};
```
kalinkaaaa14, line 12: You have already type created for it, reuse it

packages/shared/src/modules/groups/types/group-item-response-dto.type.ts
```diff
@@ -0,0 +1,10 @@
+import { type ValueOf } from "../../../libs/types/types.js";
+import { type GroupKey } from "../groups.js";
+
+type GroupResponseDto = {
```
kalinkaaaa14, line 4: GroupItemResponseDto - same as file name

packages/shared/src/modules/groups/types/group-item-dto.type.ts
kalinkaaaa14, line 1: filename has not changed

apps/backend/src/modules/groups/libs/exceptions/group-not-found.exception.ts
```diff
@@ -0,0 +1,21 @@
...
+
+class GroupError extends HTTPError {
+	public constructor({
+		cause,
+		message = GroupExceptionMessage.GROUP_NOT_FOUND,
+		status = HTTPCode.UNAUTHORIZED,
+	}: Constructor = {}) {
+		super({ cause, message, status });
+	}
+}
```
liza-veis, line 19: let's keep message and status required for the group error
```suggestion
class GroupError extends HTTPError {}
```

apps/backend/src/modules/users/libs/types/user-password-details.type.ts
```diff
@@ -1,5 +1,9 @@
+import { type GroupItemWithPermissionsDto } from "@smartscapes/shared";
```
liza-veis, line 1: should be imported from users/libs/types/types, all imports from shared should be re-exported

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -1,3 +1,5 @@
+import { type UserGetByIdItemResponseDto } from "@smartscapes/shared";
```
liza-veis, line 1: same here about shared

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -11,52 +14,183 @@ class UserRepository implements Repository {
...
 				passwordSalt,
 			})
 			.returning("*")
 			.execute();
 
-		return UserEntity.initialize(user);
+		return UserEntity.initialize({
+			email: user.email,
+			firstName: user.firstName,
+			group: user.group ? GroupEntity.initialize(user.group).toObject() : null,
```
liza-veis, line 36: Why do we have this condition on the repository level? We query the group, and it's guaranteed by the db that we will have this group

apps/backend/src/modules/users/user.service.ts
```diff
@@ -1,27 +1,39 @@
+import {
+	CommonExceptionMessage,
+	GroupKey,
+	type UserAuthResponseDto,
+	type UserGetByIdItemResponseDto,
+} from "@smartscapes/shared";
```
liza-veis, line 6: each imported entity here should be re-exported from types/types or enums/enums

apps/backend/src/modules/users/user.service.ts
```diff
@@ -33,17 +45,36 @@ class UserService implements Service {
...
+		const newUser = await this.userRepository.findById(item.toObject().id);
+
+		if (!newUser) {
+			throw new UserError({
+				message: CommonExceptionMessage.COMMON_EXCEPTION_MESSAGE,
+				status: HTTPCode.NOT_FOUND,
+			});
+		}
+
+		return newUser.toObject() as UserAuthResponseDto;
```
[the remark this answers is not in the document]
DenisDemirov: Maybe i do not understand how it must be done. But i tried different approaches to make it how you suggested before and it not working. 

- I tried apply join in user insert query, but in result i got empty fields group/permissions info in response (if i understand right this data can`t be joined during insertion)
Screenshot 2025-08-08 012846
Screenshot 2025-08-08 013132
- I tried to make transaction with create new user and return newly created user by select by id after that. But join query bricking test like it [was before](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/160#issuecomment-3161005608).

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -11,52 +14,183 @@ class UserRepository implements Repository {
...
+				"users.password_hash as passwordHash",
+				"users.password_salt as passwordSalt",
+				"users.group_id as groupId",
+			)
+			.withGraphJoined("group.permissions")
+			.where("users.email", email)
 			.first();
 
-		return user ?? null;
+		if (!user || !user.group || !user.group.permissions) {
```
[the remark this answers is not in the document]
DenisDemirov: it will caause problem with possibly undefined types
зображення

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -11,52 +14,183 @@ class UserRepository implements Repository {
...
+			return null;
+		}
+
+		return {
+			firstName: user.firstName,
+			group: {
```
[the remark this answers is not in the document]
DenisDemirov: I cant use it because i need to convert use.group.permissions items form PermissionModel[] to PermissionEntity.toObject()

apps/backend/src/modules/users/user.repository.ts
DenisDemirov: No, because we cant create user in db and got group attached in single insert query. Maybe i do not understand how it must be done? Please tell me how this query must be written? Because i found no information about how to do it

apps/backend/src/modules/users/user.service.ts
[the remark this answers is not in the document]
liza-veis: Ok, I see what the main issue is. Let's start with the unnecessary conditions for group existence even when we are sure it's present, e.g., in the getById method. It's important to differentiate between cases where something is possible by application logic and when it's just a typescript error, which can't perfectly understand the logic in the way we do.

We know the group will be present in the getById method, so this check everywhere in such places is just to satisfy TypeScript, not because it's possible or the application needs this. For these cases, we will use typecast. The reason for this is that we don't want unnecessary logic/code/checks to be present in the final build, and TypeScript types are going to be eliminated anyway. So in case we have a model with an optional group property, but from a specific query we return this group and there is no way we will get null or undefined, even if it's defined this way in the Model, we will use type casting instead of checking if the group is present and passing null if not.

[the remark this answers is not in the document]
liza-veis: BTW, let's wrap those calls in a transaction. Is there an issue with using a transaction for the tests in this case? If so, could you describe it in more details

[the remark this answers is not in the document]
DenisDemirov: > BTW, let's wrap those calls in a transaction. Is there an issue with using a transaction for the tests in this case? If so, could you describe it in more details

The problem was not transaction itself - test falled because  of WithGraphJoined in query

liza-veis:
> About test problem WithGraphJoined, it look like this. I think that our test environment can't look inside this kind of query

Yeah, we already took a look with Alina and found a way to reconcile withGraphJoined with tests. But we decided it's better to stick to a slightly different approach - instead of mapping in the models and using withGraphJoined, we decided to shift more towards using separate explicit joins in places where needed. We will update other places where we already have withGraphJoined gradually.

apps/backend/src/modules/users/user.service.ts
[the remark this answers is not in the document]
DenisDemirov: No, plain query returning me very bad and "flat" format of object
зображення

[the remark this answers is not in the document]
DenisDemirov: That how result in findAll looks like when i try to get access to group field
зображення
Screenshot 2025-08-09 200024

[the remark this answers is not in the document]
DenisDemirov: That the actual how answer from this query looks like 
зображення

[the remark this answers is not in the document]
liza-veis: I see, then it seems like this approach would require defining those fields in the model in the same style, e.g., instead of group
```ts
UserModel {
  permissionKey,
  permissionName,
  groupKey,
  groupName
}
```
But this would also require remapping and grouping later, which is not currently implemented in the app. So as I mentioned below, we already came to the solution for tests with withGraphJoined, but were considering plain join as more straightforward and beneficial. But since our structure wasn't prepared for this and we don't have a ready-made solution right now, let's switch back to withGraphJoined approach, and here is how we can fix the tests:

1. We need to pass dialect pg when we initialize knex in tests
2. We need to return from trackers the format, knex usually generates and selects 

```
databaseTracker.on.select("users").response([
      {
        ...UserEntity.initialize(mockUser).toObject(),
        "group:id": 2,
        "group:key": "users",
        "group:name": "Users",
        "group:permissions:id": 1,
        "group:permissions:key": "READ",
        "group:permissions:name": "Can read",
      },
    ]);
  });
```

let me know if this approach works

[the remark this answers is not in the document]
DenisDemirov: Its not working. Maybe i make something wrong ? I even tried to mock schema? but didn`t helped
зображення

`import knex from "knex";
import { createTracker, MockClient, type Tracker } from "knex-mock-client";
import assert from "node:assert/strict";
import { afterEach, beforeEach, describe, it } from "node:test";

import { GroupEntity } from "../groups/group.entity.js";
import { UserEntity } from "./user.entity.js";
import { UserModel } from "./user.model.js";
import { UserRepository } from "./user.repository.js";

describe("UserRepository", () => {
	let userRepository: UserRepository;
	let databaseTracker: Tracker;

	const mockGroup = GroupEntity.initialize({
		id: 2,
		key: "users",
		name: "Users",
	}).toObject();

	const mockUser = UserEntity.initialize({
		email: "test@example.com",
		firstName: "John",
		group: mockGroup,
		groupId: 2,
		id: 1,
		lastName: "Doe",
		passwordHash: "hash",
		passwordSalt: "salt",
	}).toObject();

	beforeEach(() => {
		const database = knex({ client: MockClient, dialect: "pg" });
		databaseTracker = createTracker(database);
		databaseTracker.on
			.select((builder) => {
				return builder.sql.toLowerCase().includes("information_schema.columns");
			})
			.response([]);

		databaseTracker.on
			.select((builder) => {
				return builder.sql.toLowerCase().includes("from \"users\"");
			})
			.response([
				{
					email: "test@example.com",
					firstName: "John",
					groupId: 2,
					groupKey: "users",
					groupName: "Users",
					lastName: "Doe",
					permissionId: 1,
					permissionKey: "READ",
					permissionName: "Can read",
					userId: 1,
				},
			]);

		UserModel.knex(database);
		userRepository = new UserRepository(UserModel);
	});

	afterEach(() => {
		databaseTracker.reset();
	});

	it("create create and return new user", async () => {
		const userEntity = UserEntity.initialize({
			email: "test@example.com",
			firstName: "John",
			group: mockGroup,
			groupId: 2,
			id: 1,
			lastName: "Doe",
			passwordHash: "hash",
			passwordSalt: "salt",
		});

		databaseTracker.on.insert("users").response([userEntity]);

		const result = await userRepository.create(userEntity);

		assert.deepStrictEqual(result, userEntity);
	});

	it("findAll should return all users", async () => {
		databaseTracker.on.select("users").response([
			{
				...UserEntity.initialize(mockUser).toObject(),
				"group:id": 2,
				"group:key": "users",
				"group:name": "Users",
				"group:permissions:id": 1,
				"group:permissions:key": "READ",
				"group:permissions:name": "Can read",
			},
		]);

		const result = await userRepository.findAll();

		assert.deepStrictEqual(result, [mockUser]);
	});
});
`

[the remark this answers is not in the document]
liza-veis: instead of what you have with information schema, try to use 
```ts
databaseTracker.on.any("information_schema").response([]);
```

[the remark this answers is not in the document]
DenisDemirov: Yes. Thank you - that's working

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -11,52 +16,202 @@ class UserRepository implements Repository {
...
+			});
 
-		return UserEntity.initialize(user);
+			const user = await UserModel.query()
+				.where("users.id", userId)
+				.withGraphJoined("group.permissions")
+				.first();
+
+			if (!user) {
+				throw new Error("Error");
```
liza-veis, line 38: throw a meaningful UserError here

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -11,52 +16,202 @@ class UserRepository implements Repository {
...
+							id: user.group.id,
+							key: user.group.key,
+							name: user.group.name,
+							permissions: user.group.permissions
+								? user.group.permissions.map((per) =>
+										PermissionEntity.initialize(per).toObject(),
+									)
+								: [],
+						}).toObject()
+					: null,
```
liza-veis, line 55: Aren't group and permissions always present?  It was the result we were trying to achieve, to eliminate all the unnecessary checks for group and permissions existence. The only case that needs to be checked is with permissions; in case the user won't have permissions, will it return an array or null? If it's an array, we definitely don't need this condition. 

Again, if the entities are guranteed to be returned from the db and it's just ts issue, we need to use cast and get rid of unnecessary checks

packages/shared/src/modules/permission/types/permission-item-response-dto.ts
liza-veis, line 1: file name should end with .type.ts

apps/backend/src/modules/users/user.service.ts
liza-veis, line 77: As I understood, now we are returning a user with groups and permissions from the repository
```suggestion
		return item.toObject();
```

apps/backend/src/modules/users/user.repository.ts
DenisDemirov: Yes. Sorry i forgot to change it

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -11,52 +16,199 @@ class UserRepository implements Repository {
...
-		return UserEntity.initialize(user);
+			const user = (await UserModel.query()
+				.where("users.id", userId)
+				.withGraphJoined("group.permissions")
+				.first()) as UserModel;
+
+			const group = user.group as NonNullable<typeof user.group>;
+			const permissions = user.group?.permissions as NonNullable<
+				typeof group.permissions
+			>;
```
liza-veis, line 40: this looks good let's do it the same way for all other queries in the repository

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -11,52 +16,199 @@ class UserRepository implements Repository {
...
+
+			return UserEntity.initialize({
+				email: user.email,
+				firstName: user.firstName,
+				group: GroupEntity.initializeWithPermissions({
+					id: group.id,
+					key: group.key,
+					name: group.name,
+					permissions: permissions.map((per) =>
```
liza-veis, line 49: We need to avoid contractions except for very common ones,
```suggestion
					permissions: permissions.map((permission) =>
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -11,52 +16,199 @@ class UserRepository implements Repository {
...
+		});
 	}
 
 	public async findById(id: number): Promise<null | UserEntity> {
-		const user = await this.userModel.query().findById(id);
+		const user = (await this.userModel
+			.query()
+			.findById(id)
+			.withGraphJoined("group.permissions")
+			.first()) as UserModel;
```
liza-veis, line 144: The user can be null here, so we need to return null if the user is not present, the same for findPasswordDetails. In case of create method, we know for sure that if an error wasn't thrown, we have a user, but for finding by parameters, we don't know that
```suggestion
			.first());
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -11,52 +16,199 @@ class UserRepository implements Repository {
...
+				key: group.key,
+				name: group.name,
+				permissions: permissions.map((permission) =>
+					PermissionEntity.initialize({
+						id: permission.id,
+						key: permission.key,
+						name: permission.name,
+					}).toObject(),
+				),
+			},
```
liza-veis, line 205:
```suggestion
			group: GroupEntity.initializeWithPermissions({
				id: group.id,
				key: group.key,
				name: group.name,
				permissions: permissions.map((permission) =>
					PermissionEntity.initialize(permission).toObject(),
				),
			}).toObject(),
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -11,52 +16,199 @@ class UserRepository implements Repository {
...
+			passwordHash: user.passwordHash,
+			passwordSalt: user.passwordSalt,
+		});
 	}
 
 	public async findPasswordDetails(
 		email: string,
 	): Promise<null | UserPasswordDetails> {
-		const user = await this.userModel
+		const user = (await this.userModel
```
liza-veis, line 177: In findPasswordDetails we are also not sure if the user is present and need to return null if not, be aware not to use casting  when actually the case with null is possible

liza-veis (changes requested):

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -82,7 +234,32 @@ class UserRepository implements Repository {
...
+
+		const group = updatedRow.group as NonNullable<typeof updatedRow.group>;
+		const permissions = updatedRow.group?.permissions as NonNullable<
+			typeof group.permissions
+		>;
+
+		return UserEntity.initialize({
+			email: updatedRow.email,
+			firstName: updatedRow.firstName,
+			group: GroupEntity.initializeWithPermissions({
```
liza-veis, line 249: I see we don't join user permissions in the patch query, so I believe this won't work. We need to join groups here too

lavandosya (approved):

PR #191, opened 2025-08-01
feat(frontend): toast notification smoother show ss-181

DenisDemirov:
New notification animation :
New

Old :
Old


apps/frontend/src/libs/components/toast-container/toast-container.tsx
```diff
@@ -1,22 +1,23 @@
...
 
 const ToastContainer = (): React.JSX.Element => {
 	return (
 		<LibraryToastContainer
-			autoClose={5000}
+			autoClose={3000}
 			closeOnClick
 			draggable
-			hideProgressBar={false}
+			hideProgressBar
```
kalinkaaaa14, line 12: Let's show progress bar?

PR #198, opened 2025-08-02
feat(backend/shared): search closest points to authenticated user ss-188 

vyacheslavKhoroshev:
- add abilitty to validate query parameters;
- created validation schema for POIs query parameters;
- created findNearBy method in repository;
- updated Swagger documentation;
- updated tests;

Closes №188


vyacheslavKhoroshev:
- branch 188-fear-poi-search based on 112-feat-poi-postgis;
- all commits with wrong issue‑number (118 instead 188); 

:)

vldkorin:
> - branch 188-fear-poi-search based on 112-feat-poi-postgis;
> - all commits with wrong issue‑number (118 instead 188); 
> 
> :)

Anyway, looks good

apps/backend/src/modules/points-of-interest/points-of-interest.controller.ts
```diff
@@ -214,11 +265,43 @@ class PointsOfInterestController extends BaseController {
...
+	 *                       location:
+	 *                         type: "Point"
+	 *                         coordinates: [30.5234, 50.4501]
 	 */
-	public async findAll(): Promise<
-		APIHandlerResponse<PointsOfInterestResponseDto[]>
-	> {
-		const { items } = await this.pointsOfInterestService.findAll();
+	public async findAll(
+		options?: PointsOfInterestFindAllOptions,
```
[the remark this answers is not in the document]
vyacheslavKhoroshev: > PointsOfInterestFindAllOptions

not really, I thought about it for about hour when I did it for the first time)) so I understand that it doesn't look good. I just wanted to leave the behavior where we can't pass query parameters. and I didn't want to change APIHandlerOptions type.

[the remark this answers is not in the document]
liza-veis: We will receive the options object in any case, even if the query is not passed

apps/backend/src/modules/points-of-interest/points-of-interest.repository.ts
```diff
@@ -99,6 +100,37 @@ class PointsOfInterestRepository implements Repository {
...
+			.whereRaw(
+				"ST_DWithin(location::geography, ST_SetSRID(ST_MakePoint(?, ?), 4326)::geography, ?)",
+				[longitude, latitude, radius],
+			)
+			.orderByRaw("distance ASC")
+			.execute();
+
+		return pointsOfInterest
+			.map((point) => PointsOfInterestEntity.initialize(point))
+			.filter(Boolean);
```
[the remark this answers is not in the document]
kalinkaaaa14: Not resolved yet

apps/backend/src/modules/points-of-interest/points-of-interest.service.ts
```diff
@@ -47,15 +48,28 @@ class PointsOfInterestService implements Service {
...
 
-	public async findAll(): Promise<
-		CollectionResult<PointsOfInterestResponseDto>
-	> {
-		const items = await this.pointsOfInterestRepository.findAll();
+	public async findAll(
+		query: null | PointsOfInterestSearchQuery,
+	): Promise<CollectionResult<PointsOfInterestResponseDto>> {
+		const DEFAULT_SEARCH_RADIUS_KM = 5;
+		const THOUSAND = 1000;
```
kalinkaaaa14, line 55: we need to move this variable somewhere outside (as probably such data can be reused in other places as well), and call it METERS_IN_KM

apps/backend/src/modules/points-of-interest/libs/enums/sorting-order.enum.ts
```diff
@@ -0,0 +1,6 @@
+const SortingOrder = {
```
liza-veis, line 1: let's move sorting order to common enums, outside of points of interest module

apps/backend/src/modules/points-of-interest/points-of-interest.service.ts
```diff
@@ -47,15 +48,28 @@ class PointsOfInterestService implements Service {
...
+
+		const hasLocationFilter =
+			Boolean(query?.latitude) && Boolean(query?.longitude);
+
+		const radiusMeters = (query?.radius ?? DEFAULT_SEARCH_RADIUS_KM) * THOUSAND;
+
+		const searchParameters = {
+			...query,
+			radius: radiusMeters,
+		} as PointsOfInterestSearchQuery;
```
liza-veis, line 65: why do we need type cast here, shouldn't it match the type
```suggestion
		const searchParameters: PointsOfInterestSearchQuery = {
			...query,
			radius: radiusMeters,
		};
```

packages/shared/src/modules/points-of-interest/libs/validation-schemas/points-of-interest-search-query.validation-schema.ts
```diff
@@ -0,0 +1,48 @@
...
+	})
+	.refine(
+		({ latitude, longitude, radius }) =>
+			!radius || (Boolean(latitude) && Boolean(longitude)),
+		{
+			message: PointsOfInterestValidationMessage.COORDS_REQUIRED_WITH_RADIUS,
+		},
+	)
+	.refine(
+		({ latitude, longitude }) => Boolean(latitude) === Boolean(longitude),
```
liza-veis, line 42: from logical perspective it should be and, not equality
```suggestion
		({ latitude, longitude }) => Boolean(latitude) && Boolean(longitude),
```

liza-veis (changes requested):

apps/backend/src/modules/points-of-interest/points-of-interest.service.ts
vyacheslavKhoroshev: because we don't know if there are cords or not

image

packages/shared/src/modules/points-of-interest/libs/validation-schemas/points-of-interest-search-query.validation-schema.ts
vyacheslavKhoroshev: No, after this changes I can not sent request without query.

kalinkaaaa14 (approved):

packages/shared/src/modules/points-of-interest/libs/validation-schemas/points-of-interest-search-query.validation-schema.ts
liza-veis: oh i see, I was thinking that we need both options present

PR #199, opened 2025-08-02
feat(frontend): frontend protected routes ss-24

timosh-yura:
Changes:

- added ProtectedRoutes component
- wrapped <App/> route with ProtectedRoutes


D1monB:
Should this component be like moved to a separate file ? 
Where do we need to check for user role if we want to show a specific dashboard for Users and Admins in the file with all routes or in the dashboard component itself?

It can be route /dasboard role === 'admin' ? 'AdminPanel'  : ' UsersPanel' or Factory Method in Dashboard component where we check role and return a certain component

liza-veis (changes requested):

liza-veis:
> I guess the task is blocked until then

Hm, why is it blocked? The current implementation seems to be ok for me. We only need to block requests to /app route for unauthorized users for now and allow /app/sign-in and /app/sign-up, in future we may also have /app/explore or other pages in public

timosh-yura:
Now works as intended
*added getAuthenticatedUser dispatch in <App/> component, because authUser store is also required in <Auth/>

apps/frontend/src/index.tsx
```diff
@@ -32,6 +33,10 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
 								element: <Auth />,
 								path: AppRoute.SIGN_UP,
 							},
+							{
+								element: <ProtectedRoute />,
+								path: AppRoute.APP,
+							},
```
kalinkaaaa14, line 39: You need to pass through props component which you want to render if the user is authorized. 

Currently you've made approach with Outlet and we need to get rid of it. For example, if I have a separate page which i want to cover with protection, and it didn't have any children in the route, what should I do?

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -23,14 +29,30 @@ const App = (): React.JSX.Element => {
...
+		const dispatchUser = async (): Promise<void> => {
+			try {
+				await dispatch(authActions.getAuthenticatedUser());
+			} finally {
+				setIsLoading(false);
+			}
+		};
+
+		void dispatchUser();
 	}, [isRoot, dispatch]);
```
kalinkaaaa14, line 50:
```suggestion
	useEffect(() => {
		if (hasToken) {
			dispatch(authActions.getAuthenticatedUser());
		}
	}, [dispatch, hasToken]);
```

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -23,14 +29,30 @@ const App = (): React.JSX.Element => {
...
+
+		const dispatchUser = async (): Promise<void> => {
+			try {
+				await dispatch(authActions.getAuthenticatedUser());
+			} finally {
+				setIsLoading(false);
+			}
+		};
+
+		void dispatchUser();
```
liza-veis, line 49: We need to remove this logic, and only dispatch getAuthenticatedUser as Alina mentioned, loading needs to be covered in the protected route itself

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,20 @@
...
+	);
+
+	const isAuthenticated = Boolean(authenticatedUser);
+
+	return isAuthenticated ? (
+		<Outlet />
+	) : (
+		<Navigate replace to={AppRoute.SIGN_IN} />
+	);
+};
```
liza-veis, line 18: Instead of the outlet, this route needs to accept any children (page), await the user's status while showing a loader, and then either redirect to sign-in or return the page
```suggestion
const ProtectedRoute = ({ children }: Properties): React.JSX.Element => {
	const authenticatedUser = useAppSelector(
		({ auth }) => auth.authenticatedUser,
	);
	const authDataStatus = useAppSelector(
		({ auth }) => auth.dataStatus,
	);

	const isLoading =
		dataStatus === DataStatus.PENDING || dataStatus === DataStatus.IDLE;

	if (isLoading) {
		return (
			<div className={styles["loader-container"]}>
				<Loader />
			</div>
		);
	}
	
	if (!authenticatedUser) {
		return <Navigate replace to={AppRoute.SIGN_IN} />;
	}

	return <>{children}</>;
};
```

liza-veis (changes requested):

apps/frontend/src/index.tsx
liza-veis: In order to implement this, we need to remove everything from the current App component itself and leave only getAuthenticatedUser dispatch and return outlet. And to wrap the dashboard page that we currently see in the app in a protected route, let's move the content of the app into a separate component - e.g., dashboard, and then we should be able to define it like this
```tsx
	{
		children: [
			{
				element: <Dashboard />,
				path: AppRoute.ROOT,
			},
		],
		element: <App />,
		path: AppRoute.APP,
	},
```

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -1,56 +1,41 @@
...
 	const dispatch = useAppDispatch();
-	const mockUserWithoutAvatar = {
-		avatarUrl: null,
-		firstName: "John",
-		lastName: "Smith",
-	};
 
 	const isRoot = pathname === AppRoute.APP;
 
+	const hasToken = useAppSelector(({ auth }) => auth.authenticatedUser);
```
kalinkaaaa14, line 20: you should check whether the token is stored in storage, and that's the value of hasToken

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -1,56 +1,41 @@
...
 		if (isRoot) {
 			void dispatch(userActions.loadAll());
 		}
 	}, [isRoot, dispatch]);
 
+	useEffect(() => {
+		if (!hasToken) {
+			void dispatch(authActions.getAuthenticatedUser());
+		}
+	}, [dispatch, hasToken]);
```
kalinkaaaa14, line 32: The logic should be the opposite - if token exists - fetch auth user

```suggestion
	useEffect(() => {
		if (hasToken) {
			void dispatch(authActions.getAuthenticatedUser());
		}
	}, [dispatch, hasToken]);
```

liza-veis (changes requested):

timosh-yura:
I`m not sure why, but I`m trapped in an infinite loader, if there`s no token in storage and it doesnt dispatch getAuthenticatedUser
app.tsx:
useEffect(() => {
		const fetchToken = async (): Promise<void> => {
			const hasToken = await storage.get(StorageKey.TOKEN);

			if (hasToken) {
				void dispatch(authActions.getAuthenticatedUser());
			}
		};

		void fetchToken();
	}, [dispatch]);

liza-veis:
> Im not sure why, but Im trapped in an infinite loader, if there`s no token in storage and it doesnt dispatch getAuthenticatedUser

It's because we treat the idle state for user data as loading too. So if there is no token, the authenticated user is not fetched, so the data status is idle, and it's considered pending in the app. It seems like we need to only use the pending state when showing the loader in protected routes. Overall, this idle state was added to avoid the initial switch, before we started to load the user, just to make sure it's not an issue and looks good even without this.

As for the token retrieval itself, do this inside the getAuthenticatedUser action, by our app architecture, all side effects on the frontend are executed strictly through dispatching actions, so in the app component you will only have this: 

```ts
useEffect(() => {
    void dispatch(authActions.getAuthenticatedUser());
}, [dispatch]);
```

timosh-yura:
okay, I reverted token check

liza-veis:
We need this token check, but it should happen inside loadAuthenticatedUser action creator

timosh-yura:
also token check is already in getAuthenticatedUser Thunk

kalinkaaaa14 (approved):

PR #215, opened 2025-08-04
feat(frontend): implement basic maps integration ss-116

### LINKED ISSUE #116: Implement basic maps integration (display mock data) (explore page)

ZakZh:
Add Mapbox integration into the frontend application, enabling mapping functionality with features like displaying maps, user location, and points of interest (POIs). Key changes include adding the Mapbox library, creating reusable map components, and updating the app to showcase various map types and controls.

Created Map component and reusable components for Map features, such as `CurrentPosition`, `LocationControl`, `MapControl`, and `PoiLayer`. 

Closes #116


kalinkaaaa14:
Attach a video/screenshots how it looks like please

ZakZh:
POIs with handler

https://github.com/user-attachments/assets/e599b126-1b57-461b-8af5-8b155436f17b

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -46,6 +110,56 @@ const App = (): React.JSX.Element => {
...
+							</span>
+							<Map
+								center={DEFAULT_POSITION_COORDINATES}
+								style="mapbox://styles/mapbox/streets-v12"
+								zoom={13}
+							/>
+						</div>
+						<div className={styles["map-container"]}>
+							<span className={styles["map-label"]}>Satellite view map</span>
+							<Map style="mapbox://styles/mapbox/satellite-v9" />
```
mxm-kotyk, line 133: Looks like the Map component just uses your current location if you don’t give it a position. So maybe we don’t really need the navigation module right now?

apps/frontend/src/libs/components/map/map.tsx
h4vefundieyoung, line 1: it's just huge. probably there is some way to split these things?

mxm-kotyk:
I'm not sure if it's a bug or just happening on my side, but when I reset the geolocation permission, reload the page, and allow it again, I get an error:
image

Here is the vireo of what I'm talking about:

https://github.com/user-attachments/assets/b5b533dd-ea0c-40c9-a50b-beb0537e897a

apps/frontend/src/libs/components/map/map.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+	useEffect(() => {
+		if (!mapContainer.current || mapInstance.current) {
+			return;
+		}
+
+		const token = import.meta.env["VITE_APP_MAPBOX_ACCESS_TOKEN"] as string;
+
+		if (token) {
+			(mapboxgl as unknown as { accessToken: string }).accessToken = token;
```
liza-veis, line 67: this library has a package for types, install them too

apps/frontend/src/libs/components/map/map.tsx
```diff
@@ -0,0 +1,190 @@
...
+		() => (center ? zoom : MAP_DEFAULTS.FALLBACK_ZOOM),
+		[center, zoom],
+	);
+
+	useEffect(() => {
+		if (!mapContainer.current || mapInstance.current) {
+			return;
+		}
+
+		const token = import.meta.env["VITE_APP_MAPBOX_ACCESS_TOKEN"] as string;
```
liza-veis, line 64: We don't access env variables directly in the components, especially common components. accessToken should be passed through props to the map from the top level. Let's create a MapProvider component that will attach an access token to mapboxgl (let's rename it to mapClient there) somewhere on the top level and so that we can access mapboxgl using context, and not import it everywhere. For env vars themselves, we use config package and need to pass values on the top level where we define all providers

liza-veis (changes requested):

apps/frontend/src/libs/components/map/map.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+		return (): void => {
+			resizeObserver.disconnect();
+		};
+	}, [isMapReady]);
+
+	return (
+		<div className={styles["map-container"]}>
+			<div
+				className={[styles["map"], className].filter(Boolean).join(" ")}
```
kalinkaaaa14, line 133: we have combineClassNames helper

apps/frontend/src/libs/components/map/libs/types/map-properties.type.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { type PointOfInterest } from "../components/poi-layer/libs/types/types.js";
+import {
+	type ControlPosition,
+	type LocationFoundEvent,
+	type ScaleUnit,
+} from "./shared.type.js";
+
+type MapProperties = {
+	center?: [number, number] | undefined;
```
kalinkaaaa14, line 11:
```suggestion
type GeolocationCoordinates = [number, number];
	center?: GeolocationCoordinates;
```

apps/frontend/src/libs/components/map/libs/constants/map-defaults.constant.ts
```diff
@@ -0,0 +1,34 @@
...
+const DEFAULT_LOCATION_ZOOM = 15;
+const DEFAULT_FLY_TO_DURATION = 2000;
+const DEFAULT_RESIZE_DELAY = 100;
+const DEFAULT_FALLBACK_ZOOM = 2;
+
+const MAP_DEFAULTS = {
+	CENTER: [DEFAULT_POSITION.LONGITUDE, DEFAULT_POSITION.LATITUDE] as [
+		number,
+		number,
+	],
```
kalinkaaaa14, line 12: can we get rid of type casting in this file?

apps/frontend/src/libs/components/map/libs/components/scale-control/scale-control.tsx
```diff
@@ -0,0 +1,55 @@
+import mapboxgl from "mapbox-gl";
+import { useEffect, useRef } from "react";
+
+import { SCALE_CONTROL_DEFAULTS } from "./libs/constants/constants.js";
+import { type ScaleControlProperties } from "./libs/types/types.js";
+
+const ScaleControl = ({
+	enabled = SCALE_CONTROL_DEFAULTS.ENABLED,
```
kalinkaaaa14, line 8:
```suggestion
	isEnabled = SCALE_CONTROL_DEFAULTS.ENABLED,
```

apps/frontend/src/libs/components/map/libs/components/scale-control/scale-control.tsx
```diff
@@ -0,0 +1,55 @@
+import mapboxgl from "mapbox-gl";
+import { useEffect, useRef } from "react";
+
+import { SCALE_CONTROL_DEFAULTS } from "./libs/constants/constants.js";
+import { type ScaleControlProperties } from "./libs/types/types.js";
+
+const ScaleControl = ({
```
kalinkaaaa14, line 7: This doesn't look like a usual component, as it always returns null, doesn't apply any changes to the dom itself. Maybe let's better create a custom hook useScaleControl with this encapsulated logic?

to call it like this

useScaleControl({ mapInstance, isMapReady, isEnabled, unit, position });

apps/frontend/src/libs/components/map/libs/components/scale-control/libs/constants/scale-control-defaults.constant.ts
```diff
@@ -0,0 +1,13 @@
+import {
+	type ControlPosition,
+	type ScaleUnit,
+} from "../../../../types/shared.type.js";
+
+const SCALE_CONTROL_DEFAULTS = {
+	ENABLED: true,
```
kalinkaaaa14, line 7:
```suggestion
	IS_ENABLED: true,
```

apps/frontend/src/libs/components/map/libs/components/poi-layer/poi-layer.tsx
```diff
@@ -0,0 +1,65 @@
+import mapboxgl from "mapbox-gl";
+import { useEffect, useRef } from "react";
+
+import {
+	COORDINATE_INDICES,
+	POI_LAYER_DEFAULTS,
+} from "./libs/constants/constants.js";
+import { type PoiLayerProperties } from "./libs/types/types.js";
+
+const PoiLayer = ({
```
kalinkaaaa14, line 10: Same quesiton here, can we create a custom hook for adding poi instead of treating it as a component?

apps/frontend/src/libs/components/map/libs/components/scale-control/scale-control.tsx
ZakZh: Since I’m already working on simplifying the map component (per @liza-veis’s suggestion), I don’t think we’ll actually need it anymore. But it was a good suggestion — I actually had something similar implemented locally, but I’ve already removed it(

apps/frontend/src/libs/components/map/map.tsx
h4vefundieyoung: @liza-veis as i understand - whole the communication process will be performed through the backend app api, so the client will never send requests directly to mapbox api. is there a purpose to store that token at all?

ZakZh:
> I checked the configuration files but didn't find it
> 
> Will clustering be added to the map in the future?
> 
> 
> 
> 

This task is just about adding the map, map controls, POIs, and current user location.

apps/frontend/src/libs/components/map-provider/map-provider.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+import {
+	type MapBoxGLWithToken,
+	type MapContextValue,
+	type MapProviderProperties,
+} from "./libs/types/types.js";
+
+const MapContext = createContext<MapContextValue | null>(null);
+
+const MapProvider = ({
```
liza-veis, line 13: I thought for a while, and let's try a cleaner approach similar to what we already have in the app:

1. Create a separate module similarly to how it's done for toast notifier, but for map client, e.g. map-client.module.ts

```ts
class MapClient {
  private map: mapboxgl.Map | null = null;

  private markers = new Map<string, mapboxgl.Marker>(); 

  constructor({ token }: Constructor) {
    mapboxgl.accessToken =  config.MAPBOX.ACCESS_TOKEN; // let's import it here on module level
  }

  public init(container: HTMLElement, options: /* if we need them*/) {
    this.map = new mapboxgl.Map({
      container,
      accessToken: this.token,
      ...options,
    });
  }

  public addMarker(id: string, { coordinates }: AddMarkerOptions /* coordinates should be of type PointGeometry['coordinates'] which we already have */) {
    if (!this.map) {
       return;
    }
    // remove old if exists
    this.removeMarker(id);

    const marker = new mapboxgl.Marker().setLngLat(coordinates).addTo(this.map);

    this.markers.set(id, marker);
  }

 public removeMarker(id: string) {
    const marker = this.markers.get(id);

    if (marker) {
      marker.remove();
      this.markers.delete(id);
    }
  }

  public destroy() {
    this.markers.forEach(marker => m.remove());
    this.map?.remove();
    this.markers.clear();
    this.map = null;
  }

  public setCenter(coordinates: PointGeometry['coordinates']) { // example, don't add if not needed
    this.instance.setCenter(coordinates);
  }

  // whatever else we need
}
```

2. Now, for the MapProvider we want to create a new instance of the map client per map (this MapProvider won't be global, but per specific map). What we want to have in this MapProvider component is something like this:

```tsx
/* Context in this file*/

const MapContext = createContext<MapClient | null>(null);

/* Hook to use this context in common hooks, I'm writing it in the middle not to make it more clear 😅  */
const useMapClient =(): MapService => {
  const client = useContext(MapContext);

  if (!client) {
     throw new MapClientError('useMapClient must be inside a MapProvider'); // Our custom error, but it will be an internal error, not a client error, so it's ok to throw it this way. This error should be defined in the map-client module exceptions
  }

  return client;
}

/* We may split map and provider (only client logic), but let's actually have them as a single component */

const Map = ({
 children,
 markers: [],
 style = 'mapbox://styles/mapbox/streets-v11', // just an example, don't add if not needed
 center = [0, 0],  // just an example, don't add if not needed
}) => {
 const clientRef = useRef<MapClient>(new MapService());
 const containerRef = useRef<HTMLDivElement>(null);


 useEffect(() => {
    const client = clientRef.current;
    const container = containerRef.current;

    if (!container) {
      return;
    }

    client.init(container, { center, style  } /* only if we need this*/);

    return () => {
      client.destroy();
    };
  }, []);


  useEffect(() => {
    for (const marker of markers) {
       const { id, coordinates } = marker;

      client.addMarker(id, { coordinates });
    }
  }, [client, markers]);


 /* the following useEffects are examples of how we would react to options change in provider*/
  useEffect(() => (clientRef.current as MapClient).setOptions(center), [center]);

  useEffect(() => (clientRef.current as MapClient).setStyle(style), [style]);

return (
    <MapContext.Provider value={clientRef.current!}>
      {/* This div is where Mapbox will render */}
      <div ref={containerRef} className={styles.map /* e.g. { width: '100%', height: '100%' } */} />
      {children}
    </MapContext.Provider>
  );
};
```
3. Usage in the app. Each map will be wrapped in a map provider. Here is how example of a full-screen app that may look like in a simplified form, just so you understand the idea (it should be adjusted to the app conventions)

```tsx
<div style={{ width: '100vw', height: '100vh' }}>
      <MapProvider markers={markers }center={[30,50]}>
        {/* UI overlayed on top of map, or if we need anything else, it would be added like this */}
        <PoiPanel />
        <div style={{ position: 'absolute', top: 10, left: 10 }}>
          <AddPoiButton />
        </div>
      </MapProvider>
    </div>

const AddPoiButton = () => {
  const dispatch = useDispatch();
  const map = useMapClient(); // MapClient from context

  const handleAdd = useCallback(() => {
    const center = map.getCenter();

    const id = crypto.randomUUID();

    dispatch(
      addPoi({
        id,
        coordinates,
      })
    );

    map.flyTo({ center: coordinates });
  }, [dispatch, map]);

  return <button onClick={handleAdd}>Add POI</button>;
}
```

liza-veis (changes requested):

apps/frontend/src/libs/modules/map-client/map-client.module.ts
```diff
@@ -0,0 +1,322 @@
...
+
+	public constructor({ token }: Constructor = {}) {
+		(mapboxgl as MapBoxGLWithToken).accessToken =
+			token || config.ENV.MAPBOX.ACCESS_TOKEN;
+
+		this.defaultMapOptions = DEFAULT_MAP_OPTIONS;
+		this.defaultMarkerOptions = DEFAULT_MARKER_OPTIONS;
+		this.defaultNavigationControlOptions = DEFAULT_NAVIGATION_CONTROL_OPTIONS;
+		this.defaultScaleControlOptions = DEFAULT_SCALE_CONTROL_OPTIONS;
+		this.defaultGeolocateControlOptions = DEFAULT_GEOLOCATE_CONTROL_OPTIONS;
```
liza-veis, line 61: We don't need to store default values in the class itself. If this is a changeable property, e.g., mapOptions, and we need to keep track of it, then it's ok, but there is no reason to store the default because the name suggests it won't change in our case
```suggestion
```

apps/frontend/src/libs/modules/map-client/map-client.module.ts
```diff
@@ -0,0 +1,322 @@
...
+		id: string,
+		coordinates: PointGeometry["coordinates"],
+	) => void;
+	private markers = new Map<string, Marker>();
+	private resizeObserver: null | ResizeObserver = null;
+	private shouldCenterOnUser: boolean = false;
+
+	public constructor({ token }: Constructor = {}) {
+		(mapboxgl as MapBoxGLWithToken).accessToken =
+			token || config.ENV.MAPBOX.ACCESS_TOKEN;
```
liza-veis, line 55: Not sure why we need typecast here, shouldn't Mapbox already have accessToken in types. Also, we don't need many ways to pass the token, only one
```suggestion
	public constructor() {
		mapboxgl.accessToken = config.ENV.MAPBOX.ACCESS_TOKEN;
```

apps/frontend/src/libs/modules/map-client/map-client.module.ts
```diff
@@ -0,0 +1,322 @@
...
+				center,
+				essential: true,
+				zoom: zoom ?? this.map.getZoom(),
+			});
+		}
+	}
+
+	public getMap(): mapboxgl.Map | null {
+		return this.map;
+	}
```
liza-veis, line 238: We don't need this for now, if we will need it, we will add a public getter, but for now let's keep it incapsulated
```suggestion
```

apps/frontend/src/libs/modules/map-client/map-client.module.ts
```diff
@@ -0,0 +1,322 @@
...
+	type MapBoxGLWithToken,
+	type MapOptions,
+	type Marker,
+	type MarkerOptions,
+	type NavigationControlOptions,
+	type PointGeometry,
+	type ScaleControlOptions,
+} from "./libs/types/types.js";
+
+import "mapbox-gl/dist/mapbox-gl.css";
```
liza-veis, line 28: Is this the recommended way by official docs 🤔

apps/frontend/src/libs/modules/map-client/map-client.module.ts
```diff
@@ -0,0 +1,322 @@
...
+		this.defaultNavigationControlOptions = DEFAULT_NAVIGATION_CONTROL_OPTIONS;
+		this.defaultScaleControlOptions = DEFAULT_SCALE_CONTROL_OPTIONS;
+		this.defaultGeolocateControlOptions = DEFAULT_GEOLOCATE_CONTROL_OPTIONS;
+	}
+
+	public addAllMapControls(): void {
+		this.addNavigationControl();
+		this.addScaleControl();
+		this.addGeolocateControl();
+	}
```
liza-veis, line 68: Too many options and conditions in all the methods, even if they are not used or not passed (not only those 3, but overall). Remove everything that is not needed or used for now and unnecessary options. We need to keep as less options for customizing something as possible. We only need one static (default) option for everything right now. 

Also, we need to keep the public interface as less as possible; we are not writing a customizable library with many ways of interacting with the things. In the actual application code, all modules should be as restricted and controlled as possible. For example, if we have addNavigationControl and other controls in public, we don't need addAllMapControls, because we will have 2 ways to achieve the same result. In case we want to keep addMapControls, then we need to make internal methods private. 

It's very hard to predict what will be used and what will be the most convenient way of that in the future. It's much easier to have a single option right now and add what's necessary when we are implementing a specific feature, rather than considering all cases right now. Most of this will be either not used or significantly changed, which is very bad for code maintenance

apps/frontend/src/pages/explore/components/explore-routes/explore-routes.tsx
```diff
@@ -0,0 +1,178 @@
+import React, { useCallback, useEffect, useState } from "react";
+import { useWatch } from "react-hook-form";
+
+import { Button, Input, Select } from "~/libs/components/components.js";
+import { type SelectOption } from "~/libs/components/select/libs/types/types.js";
+import { useAppForm, useMapClient } from "~/libs/hooks/hooks.js";
+
+import styles from "./styles.module.css";
+
+const DEFAULT_ZOOM = 15;
```
liza-veis, line 10: I think it can be just the default option of flyTo, which we want to change in the app, most probably, so let's define it there for now

apps/frontend/src/pages/explore/explore.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const Explore = (): React.JSX.Element => {
+	return (
+		<main className={styles["main"]}>
+			<MapProvider>
+				<ExploreRoutes />
+			</MapProvider>
+		</main>
+	);
+};
```
liza-veis, line 16: This may look good for the demo, but it adds way too much unnecessary logic, which we don't need, so let's remove it, and as for markers,  let's make fly to markers a default behaviour on click. In case we need to customize it, we will customize it later. Except for maintenance, another huge issue with adding unnecessary logic is that it still requires review and time for refining, while it doesn't bring value to the app

```suggestion
const Explore = (): React.JSX.Element => {
	return (
		<main className={styles["main"]}>
			<MapProvider markers={mockMarkers} />
		</main>
	);
};
```

liza-veis (changes requested):

apps/frontend/vite.config.ts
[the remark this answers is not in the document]
liza-veis: Do we encounter a specific issue with the build that requires this configuration?

apps/frontend/src/libs/components/map/map.tsx
liza-veis: missed this question. Yes, we will be using it on the client too, but not for its logic APIs, but for rendering the map itself. We can choose whatever provider we want for the client, and it could be even Google Maps or whatever, and it's not related to the backend and how it's implemented there. So even though we decided to use Mapbox for both the backend logic side and for the frontend map rendering, they shouldn't be dependent on each other. Backend should reply with a common response and shouldn't adjust it because frontend uses the Mapbox api. The same is for the frontend, it shouldn't know that the backend uses the mapbox api under the hood, and it shouldn't be exposed in any way from the backend. It's possible when you use some kind of server components or ssr, or even implement your own maps logic fully on the backend, where it will return tiles and client parts to the frontend, but this adds unnecessary complexity, is hard to implement, and there is no need for this mostly

apps/frontend/vite.config.ts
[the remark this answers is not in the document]
ZakZh: Yes — when mapbox-gl is bundled into the main (index) chunk, it makes the chunk too large, which causes the build to fail.
Before:
Знімок екрана 2025-08-08 225458

By manually splitting it using manualChunks in vite.config.ts, we force Vite to put mapbox-gl in a separate chunk, which resolves the issue.
After:
Знімок екрана 2025-08-09 205514

apps/frontend/src/libs/modules/map-client/map-client.module.ts
ZakZh: Yes — this is recommended and ensures that the default Mapbox controls are styled correctly

apps/frontend/src/libs/components/map-provider/map-provider.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+	useEffect(() => {
+		const client = mapClientInstance;
+
+		if (!initializedReference.current || !client.getMap() || !center) {
+			return;
+		}
+
+		client.flyTo(center);
+	}, [center, mapClientInstance]);
```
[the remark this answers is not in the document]
ZakZh: i handle it in module level. When there is no center it just center on current user position, but because of TS error i just leave here some simple !center check

apps/frontend/src/libs/modules/map-client/map-client.module.ts
ZakZh: About that...
@types/mapbox is now integrated into the mapbox-gl package itself. However, the core library doesn't explicitly expose the accessToken type. Instead, it's included as part of the MapOptions type — so I extracted and reused it from there to maintain correct typing.
Знімок екрана 2025-08-09 223201

apps/frontend/src/libs/modules/map-client/map-client.module.ts
ZakZh: init – Initializes the map. 
addAllMapControls – Adds all standard controls at default positions. 
addNavigationControl, addScaleControl, addGeolocateControl – Add specific controls with optional config/position. 
addMarker – Adds a POI marker to the map. 
addMarkers – Batch add POIs. 
removeMarker clearAllMarkers – Remove single or all markers. 
setMarkerClickHandler – Attach click handler for POIs. 
flyTo, setCenter – Move the map programmatically. 
resize – Resize the map container on layout changes. 
destroy – Cleanup all resources and detach map. 

All methods in this class, I think, required for the target functionality.
But let me know if any part should be deleted, but currently it feels scoped to the task - add map, controls, POIs.

liza-veis: addAllMapControls -> let's remove it and call addNavigationControl, addScaleControl, addGeolocateControl in the map provider in use effect on initialization

addMarker - let's make it private and expose only addMarkers to the public interface 

removeMarker - the same, I think we only need it as a private method for now

 setMarkerClickHandler - Let's remove it. We will implement it later when we add synchronization, as it's not needed for now. Let's make flyTo the marker the default behavior inside mapClient. If we need any customization, we will implement it later when we specifically know what we need, and what way of implementation will be more convenient for us, because for now we don't want to predict the convenient way and to enlarge this pr

Other methods are ok. And by this comment, I not only meant methods, but also unnecessary options. For example, in addNavigationControl, we accept options and merge them with default while we don't need this for now and only use the default options and the same for mapOptions, or other methods, that we don't pass for now

liza-veis: We also don't actually need setMarkerClickHandler, which adds the ability to update the handler dynamically. For now, we can just pass onMarkerClick to the map provider and accept here on init in the options, and we will set it during the addMarker phase. We are exposing the map client for dynamic usage, but most of the needed functionality right now can be covered with just the options passed. So maybe we will get rid of the context in the future at all, and there was no need for it

liza-veis (changes requested):

apps/frontend/src/libs/modules/map-client/map-client.module.ts
liza-veis: oh, i see, ok then let's leave it as it is

apps/frontend/src/libs/modules/map-client/libs/types/types.ts
```diff
@@ -0,0 +1,53 @@
...
+	position?: ControlPosition;
+};
+
+type IControl = mapboxgl.IControl;
+
+type MapBoxGL = typeof mapboxgl;
+
+type MapBoxGLWithToken = MapBoxGL & {
+	accessToken: AccessToken;
+};
```
liza-veis, line 19: It seems like Mapbox wants us to pass accessToken, not to the mapboxgl, but rather in the map options, in this case we won't need this types
```suggestion
```

apps/frontend/src/libs/modules/map-client/libs/types/types.ts
```diff
@@ -0,0 +1,53 @@
+import { type PointGeometry } from "@smartscapes/shared";
+import { type default as mapboxgl } from "mapbox-gl";
+
+type AccessToken = MapOptions["accessToken"];
+
+type ControlPosition = mapboxgl.ControlPosition;
```
liza-veis, line 6:
```suggestion
type MapControlPosition = mapboxgl.ControlPosition;
```

apps/frontend/src/libs/modules/map-client/libs/types/types.ts
```diff
@@ -0,0 +1,53 @@
...
+
+type GeolocateControlOptions = mapboxgl.GeolocateControlOptions & {
```
liza-veis, line 8: 1. Let's define all the properties explicitly, it's ok use mapboxgl.GeolocateControlOptions type for internal implementation, but if we are going to allow passing options in such a format from the outside to other components, that's not what we want. So we either need to have a separate type for internal config and external config, or just let's define only properties that are used right now. And the same for other types here that extend mapbox types
2. Haven't noticed this previous time, but we also need to split all types into the respective files - name.type.ts
```suggestion
type GeolocateControlOptions = {
```

liza-veis (changes requested):

apps/frontend/src/libs/modules/map-client/map-client.module.ts
```diff
@@ -0,0 +1,263 @@
...
+		const { coordinates, ...markerOptions } = options;
+
+		const marker = new mapboxgl.Marker({
+			...DEFAULT_MARKER_OPTIONS,
+			...markerOptions,
+		})
+			.setLngLat(coordinates)
+			.addTo(this.map);
+
+		(marker as Marker).id = id;
```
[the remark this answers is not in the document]
liza-veis: Maybe we don't need to pass the ID to the marker, actually, we already have a map where we can retrieve it by ID

liza-veis:
and these two comments also need to be resolved 🐱 
https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/215#discussion_r2265241566
https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/215#discussion_r2265250769

liza-veis (approved):
LGTM, good job 😎

PR #217, opened 2025-08-04
feat(frontend): add custom select component ss-192

### LINKED ISSUE #192: select component with multi options

vldkorin:
Added custom select component using react-select with options:
- single select
- multi select
image
image
image

Closes #192


apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,62 @@
...
+	label,
+	name,
+	options,
+	placeholder = "",
+}: Properties<TFieldValues, TOptionValue>): React.JSX.Element => {
+	const { field } = useFormController<TFieldValues>({ control, name });
+
+	const rawValue = field.value as null | TOptionValue | TOptionValue[];
+	const value = mapSelectValue<TOptionValue>(rawValue, options, isMulti);
+	const handleChange = React.useMemo(
```
liza-veis, line 40: import useMemo from ~/libs/hooks/hooks

apps/frontend/src/libs/components/select/libs/mappers/select-value.mapper.ts
liza-veis, line 1: let's move it to helpers, we won't have mappers as separate entities, only map functions - which are helpers, or maps - which are objects that map values. also file name should match exported function name -> map-select-value.helper.ts

apps/frontend/src/libs/components/select/libs/mappers/select-value.mapper.ts
```diff
@@ -0,0 +1,19 @@
+import { type SelectOption } from "../types/select-option.type.js";
```
liza-veis, line 1:
```suggestion
import { type SelectOption } from "../types/types.js";
```

apps/frontend/src/libs/components/select/libs/mappers/select-value.mapper.ts
```diff
@@ -0,0 +1,19 @@
...
+	options: SelectOption<TOptionValue>[],
+	isMulti: boolean,
+): SelectOption<TOptionValue>[] => {
+	if (isMulti) {
+		return Array.isArray(rawValue)
+			? options.filter((option) => rawValue.includes(option.value))
+			: [];
+	}
+
+	const found = options.find((option) => option.value === rawValue);
```
liza-veis, line 14: let's make it more clear
```suggestion
	const foundOption = options.find((option) => option.value === rawValue);
```

apps/frontend/src/libs/components/select/libs/helpers/is-multi-value/is-multi-value.helper.ts
```diff
@@ -0,0 +1,13 @@
+import { type MultiValue, type SingleValue } from "react-select";
+
+import { type SelectOption } from "../../types/select-option.type.js";
+
+const isMultiValue = <TOptionValue>(
```
liza-veis, line 5: I think this helper is too trivial for a separate helper, and since we are not creating library let's get rid of it 😅

apps/frontend/src/libs/components/select/libs/helpers/select/select.helper.ts
```diff
@@ -0,0 +1,24 @@
...
+			| SingleValue<SelectOption<TOptionValue>>,
+	): void => {
+		if (isMultiValue<TOptionValue>(newValue)) {
+			const values = newValue.map((o) => o.value);
+			onChange(values);
+		} else {
+			const single = newValue?.value ?? null;
+			onChange(single);
+		}
+	};
```
liza-veis, line 22: It's better to simplify logic here to return a value, no need to call the helper itself
```suggestion
const getSelectNewValue =
	<TOptionValue>(
		onChange: (value: null | TOptionValue | TOptionValue[]) => void,
	) => MultiValue<SelectOption<TOptionValue>>
			| SingleValue<SelectOption<TOptionValue>>=> {
	   return Array.isArray(newValue) ? newValue.map((option) => option.value) : newValue?.value ?? null;
	};
```

apps/frontend/src/libs/components/select/libs/constants/select-class-names.constant.ts
```diff
@@ -0,0 +1,27 @@
+import { combineClassNames } from "~/libs/helpers/helpers.js";
+
+import styles from "../../styles.module.css";
+
+const selectClassNames = {
```
liza-veis, line 5: By conventions in style guide we need to name objects in the singular form, don't forget to update file name
```suggestion
const selectStylesConfig = {
```

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,62 @@
...
+	isMulti = false,
+	label,
+	name,
+	options,
+	placeholder = "",
+}: Properties<TFieldValues, TOptionValue>): React.JSX.Element => {
+	const { field } = useFormController<TFieldValues>({ control, name });
+
+	const rawValue = field.value as null | TOptionValue | TOptionValue[];
+	const value = mapSelectValue<TOptionValue>(rawValue, options, isMulti);
```
liza-veis, line 39: use use memo for this to, not to compute it on fly

liza-veis (changes requested):
The appearance looks cool 😎, a few comments

kalinkaaaa14 (approved):

liza-veis (approved):

PR #218, opened 2025-08-04
feat(frontend/backend/shared): profile page ss-185

### LINKED ISSUE #185: feat: Add my profile page with editing
- [x] Add new page My Profile on UI. Add this as a separate option on the sidebar
- [x] Add two tabs - User Details, and Photos. For now, Photos would be empty block
- [x] It should be similar to this design, but in our color schema (main color is green instead of brown, buttons are not rounded)
- [x] Component with updating avatar should be added, but without any logic. Just a markup
- [x] Implement the ability to update the user on the backend

Image

vyacheslavKhoroshev:
**Backend**

- add patch method to user module;
- updated tests;
- upadated swagger documentation;
- fix sign-in returned data;

**Frontend**

- add patch method to user API;
- create tab-container component;
- profile page with adaptive layout;
- fixed the stored authenticatedUser data after sign-in;
- fixed import of user types;

Closes #185


apps/frontend/src/pages/profile/libs/components/user-details/libs/components/profile-form/styles.module.css
```diff
@@ -0,0 +1,21 @@
+.form {
+	display: flex;
+	flex-direction: column;
+	gap: 30px;
+}
+
+.form > button {
```
h4vefundieyoung, line 7: we can't use element selectors according to the quality criteria

apps/frontend/src/pages/profile/libs/components/user-details/libs/components/profile-form/styles.module.css
```diff
@@ -0,0 +1,21 @@
...
+	width: 50%;
+}
+
+@media screen and (width <= 768px) {
+	.form > button {
```
h4vefundieyoung, line 12: same

h4vefundieyoung: and i believe you will find more there, aint gonna mark everything

apps/frontend/src/libs/components/tab-container/tab-container.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const TabContainer = ({
+	activeTabClassName,
+	containerClassName,
+	contentClassName,
+	defaultTabId,
+	navigationClassName,
+	tabClassName,
+	tabs,
+}: TabContainerProperties): React.JSX.Element => {
```
liza-veis, line 15: We need to avoid passing classnames to common ui components (that already have predefined styles), to have all possible combinations/looks fixed in the component itself, and avoid unexpected side-effects across the app.  If any adjustment is needed, e.g., small/big, style - primary/secondary - it should be done explicitly through props and fixed interface. In our case, I believe we don't need all of this, and we only need items, while everything else can be defined here as part of the common logic
```suggestion
const Tabs = ({
    activeTabId,
	items,
}: Properties): React.JSX.Element => {
```

We also need to create a hook, something like useTabsNavigation, which will be used outside of this component, and set appropriate query params/id to url and detect which tab is currently active by URL. This hook will return activeTabId or something like this, which we will pass to this component

apps/frontend/src/libs/components/tab-container/libs/types/tab-container-properties.type.ts
```diff
@@ -0,0 +1,17 @@
+type Tab = {
+	element: React.JSX.Element;
+	id: string;
+	label: string;
+};
+
+type TabContainerProperties = {
```
liza-veis, line 7: let's keep this in component as Properties, no need to move it separately because it's not reused, the only thing we need separately is TabItem

liza-veis (changes requested):

packages/shared/src/modules/users/libs/validation-schemas/user-profile.validation.schema.ts
```diff
@@ -0,0 +1,16 @@
...
+import { lastNameValidationSchema } from "./last-name.validation-schema.js";
+
+const userProfile = z
+	.object({
+		firstName: firstNameValidationSchema.optional(),
+		lastName: lastNameValidationSchema.optional(),
+	})
+	.refine((data) => data.firstName || data.lastName, {
+		message: UserValidationMessage.FIRST_OR_LAST_NAME_REQUIRED,
+	});
```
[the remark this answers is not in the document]
vyacheslavKhoroshev: I do not get it (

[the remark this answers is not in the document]
liza-veis: the same thing as for pointOfInterestUpdate - REQUIRED_FIELDS_FOR_UPDATE, I mean just to check that at least one key is present, not to least first or last name specifically

apps/frontend/src/modules/auth/slices/auth.slice.ts
```diff
@@ -56,7 +59,14 @@ const { actions, name, reducer } = createSlice({
...
 	name: "auth",
-	reducers: {},
+	reducers: {
+		setAuthenticatedUser(
+			state,
+			action: PayloadAction<UserGetByIdItemResponseDto>,
+		) {
+			state.authenticatedUser = action.payload;
+		},
+	},
```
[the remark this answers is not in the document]
vyacheslavKhoroshev: Or after updating I have to call authController.getAuthenticatedUser ?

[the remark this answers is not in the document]
liza-veis: oh yes, you are right here. But we don't need a separate simple action for this, let's just do it on fullfied action on patch user profile

packages/shared/src/modules/users/libs/validation-schemas/user-profile-patch.validation.schema.ts
```diff
@@ -0,0 +1,46 @@
...
+	.object({
+		firstName: z
+			.string()
+			.trim()
+			.min(UserValidationRule.REQUIRED_STRING_MIN_LENGTH, {
+				error: UserValidationMessage.FIRST_NAME_REQUIRED,
+			})
+			.min(UserValidationRule.FIRST_NAME_MINIMUM_LENGTH, {
+				error: UserValidationMessage.FIRST_NAME_MINIMUM_LENGTH,
+			})
```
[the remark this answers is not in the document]
vyacheslavKhoroshev: https://github.com/user-attachments/assets/e79a028d-8b72-438a-ae2b-f71f1ef8eff7

packages/shared/src/modules/users/libs/validation-schemas/user-profile-patch.validation.schema.ts
```diff
@@ -0,0 +1,46 @@
...
+			.optional(),
+		lastName: z
+			.string()
+			.trim()
+			.min(UserValidationRule.REQUIRED_STRING_MIN_LENGTH, {
+				error: UserValidationMessage.LAST_NAME_REQUIRED,
+			})
+			.min(UserValidationRule.LAST_NAME_MINIMUM_LENGTH, {
+				error: UserValidationMessage.LAST_NAME_MINIMUM_LENGTH,
+			})
```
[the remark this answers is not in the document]
vyacheslavKhoroshev: I just copied it as it was.

apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -0,0 +1,32 @@
...
+	},
+	{
+		element: <Photos />,
+		id: "photos",
+		label: "Photos",
+	},
+];
+
+const Profile = (): React.JSX.Element => {
+	const handleDeleteAccount = useCallback(() => {}, []);
```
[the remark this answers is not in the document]
vyacheslavKhoroshev: also deleted button "Delete acount"

apps/frontend/src/modules/users/slices/users.ts
```diff
@@ -1,9 +1,10 @@
-import { loadAll } from "./actions.js";
+import { loadAll, patchProfile } from "./actions.js";
 import { actions } from "./users.slice.js";
 
 const allActions = {
 	...actions,
 	loadAll,
+	patchProfile,
```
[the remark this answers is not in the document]
liza-veis: Yes agree, since we will be updating authenticated-user let's keep it in auth slice

packages/shared/src/modules/users/libs/validation-schemas/user-profile-patch.validation.schema.ts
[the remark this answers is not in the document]
liza-veis: @kalinkaaaa14 do you mean check that at least one character is required? If so, there is no other way in zod to check that field value is non empty, so it's more the standard approach to check that at least something is present

apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -0,0 +1,32 @@
+import { Button, TabContainer } from "~/libs/components/components.js";
+import { useCallback } from "~/libs/hooks/hooks.js";
+
+import { Photos, UserDetails } from "./libs/components/components.js";
+import styles from "./styles.module.css";
+
+const PROFILE_TABS = [
```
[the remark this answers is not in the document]
liza-veis: hm, I agree we need to move it to constants.tsx, however, let's keep it on profile level, e.g. `profile/libs/constants/profile-tabs.constant.ts`. As for typing, we need a common type - TabItem, similarly to NavigationItem and we need to use it here like `const PROFILE_TABS: TabItem[] = [`

packages/shared/src/modules/users/libs/validation-schemas/user-profile-patch.validation.schema.ts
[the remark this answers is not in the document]
kalinkaaaa14: Oh, ok, than let's leave it

liza-veis (changes requested):

apps/frontend/src/pages/profile/libs/components/user-details/user-details.tsx
```diff
@@ -0,0 +1,74 @@
...
+		(state) => state.auth,
+	);
+	const isLoading = dataStatus === DataStatus.PENDING;
+
+	const handleFormSubmit = useCallback(
+		(payload: UserAuthPatchRequestDto): void => {
+			if (authenticatedUser) {
+				void dispatch(
+					authActions.authPatch({
+						id: authenticatedUser.id.toString(),
```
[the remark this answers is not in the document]
vyacheslavKhoroshev: no, I need to convert to string because getFuulEndPoint expects a "string" 
image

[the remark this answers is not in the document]
vyacheslavKhoroshev: image i did this, or we could change the logic of getFuulEndPoint where we'd convert all numbers to strings....

vyacheslavKhoroshev:
https://github.com/user-attachments/assets/df9d32f4-6643-4eee-8fa5-3e940d98e363

apps/frontend/src/pages/profile/libs/components/user-details/user-details.tsx
```diff
@@ -0,0 +1,59 @@
...
+import styles from "./styles.module.css";
+
+const UserDetails = (): null | React.JSX.Element => {
+	const dispatch = useAppDispatch();
+	const { authenticatedUser, dataPatchStatus } = useAppSelector(
+		(state) => state.auth,
+	) as {
+		authenticatedUser: UserAuthResponseDto;
+		dataPatchStatus: ValueOf<typeof DataStatus>;
+	};
```
liza-veis, line 25: Select each property separately. We will access properties by the closest key, the way they are stored in the store (the same way we defined them in slice, not a specific slice as a whole), not to break memoization.

apps/frontend/src/pages/profile/styles.module.css
```diff
@@ -0,0 +1,21 @@
+.profile-container {
+	display: flex;
+	flex: 1;
+	flex-direction: column;
+	padding: 26px;
```
liza-veis, line 5: should be dividable by 4
```suggestion
	padding: 24px;
```

packages/shared/src/modules/users/libs/types/types.ts
```diff
@@ -1,3 +1,5 @@
+export { type UserAuthPatchRequestDto } from "./user-auth-patch-request-dto.type.ts.js";
+export { type UserAuthPatchResponseDto } from "./user-auth-patch-response-dto.js";
```
liza-veis, line 2:
```suggestion
export { type UserAuthPatchResponseDto } from "./user-auth-patch-response-dto.type.js";
```

packages/shared/src/modules/users/libs/validation-schemas/user-auth-patch.validation.schema.ts
```diff
@@ -0,0 +1,66 @@
...
+				(Boolean(data.lastName) && data.lastName !== originalData.lastName);
+
+			return hasChanges;
+		},
+		{
+			message: UserValidationMessage.NO_CHANGES,
+			path: ["firstName"],
+		},
+	);
+};
```
liza-veis, line 64: Let's remove this; it's not a data validation issue. react-hook-form provides properties to check if something was changed, but in our case, I think for now we don't need to account for this


```suggestion
```

packages/shared/src/modules/users/libs/validation-schemas/validation-schemas.ts
```diff
@@ -1,2 +1,3 @@
+export { authenticatedUserPatch } from "./user-auth-patch.validation.schema.js";
```
kalinkaaaa14, line 1:
```suggestion
export { authenticatedUserPatch } from "./authenticated-user-patch.validation.schema.js";
```

packages/shared/src/modules/users/users.ts
```diff
@@ -1,5 +1,7 @@
 export { UserExceptionMessage, UsersApiPath } from "./libs/enums/enums.js";
 export {
+	type UserAuthPatchRequestDto,
```
kalinkaaaa14, line 3: type AuthenticatedUserPatchRequestDto

packages/shared/src/modules/users/users.ts
```diff
@@ -1,5 +1,7 @@
...
+	type UserAuthPatchResponseDto,
```
kalinkaaaa14, line 4:
```suggestion
	type AuthenticatedUserPatchResponseDto,
```

packages/shared/src/modules/users/libs/types/user-auth-patch-response-dto.type.ts
```diff
@@ -0,0 +1,8 @@
+type UserAuthPatchResponseDto = {
```
kalinkaaaa14, line 1: I see Liza requested to rename everywhere UserAuthPatch to AuthenticatedUserPatch, so change it please everywhere, and don't forget about file names

packages/shared/src/modules/users/libs/enums/user-exception-message.enum.ts
```diff
@@ -1,6 +1,7 @@
 const UserExceptionMessage = {
 	INVALID_CREDENTIALS: "Invalid credentials",
 	USER_ALLREADY_EXISTS: "A user with this email already exists",
+	USER_NOT_FOUND: "User not found",
```
kalinkaaaa14, line 4:
```suggestion
	USER_NOT_FOUND: "User is not found",
```

apps/frontend/src/libs/components/tab-container/libs/types/tab-item.type.ts
```diff
@@ -0,0 +1,9 @@
+import { type ReactNode } from "react";
+
+type TabItem = {
+	element: ReactNode;
+	id: string;
```
kalinkaaaa14, line 5: type TabItemKey = 'user-details' | 'my-reviews';

```suggestion
	id: TabItemKey;
```

apps/frontend/src/libs/components/avatar/avatar.tsx
```diff
@@ -1,22 +1,20 @@
...
 
 import styles from "./styles.module.css";
 
 const DEFAULT_AVATAR_SIZE = 32;
 
 type Properties = {
-	avatarUrl?: string;
 	size?: number;
-	user: UserAuthResponseDto;
+	user: UserAuthResponseDto & { avatarUrl?: null | string };
```
kalinkaaaa14, line 10: Let's add avatarUrl to UserAuthResponseDto, and it should be either string or null, without undefined

apps/frontend/src/libs/hooks/use-tab-navigation/use-tab-navigation.ts
kalinkaaaa14, line 1: use-tab-navigation.hook.ts

liza-veis (approved):
LGTM, please resolve Alina's comments above 🐱

apps/frontend/src/libs/components/avatar/avatar.tsx
liza-veis: Hm, I think it's better to leave it as it is right now, since we don't have this field on the backend, and UserAuthResponseDto is shared among many services. Let's get rid of this, when we actually will return avatarUrl from the backend

packages/shared/src/modules/users/libs/validation-schemas/user-auth-patch.validation.schema.ts
```diff
@@ -0,0 +1,46 @@
...
+				error: UserValidationMessage.LAST_NAME_MAXIMUM_LENGTH,
+			})
+			.refine(checkIsLatinLetter, {
+				message: UserValidationMessage.LAST_NAME_INVALID,
+			})
+			.optional(),
+	})
+	.refine((data) => data.firstName || data.lastName, {
+		message: UserValidationMessage.FIRST_OR_LAST_NAME_REQUIRED,
+	});
```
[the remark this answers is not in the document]
liza-veis: I mean the same way as we did for points-of-interest REQUIRED_FIELDS_FOR_UPDATE. So that we don't have to check data.firstName, data.lastName, but rather that at least one key is present

apps/frontend/src/libs/components/avatar/avatar.tsx
kalinkaaaa14: Ok, let's do it in this way)

liza-veis (approved):

PR #223, opened 2025-08-05
feat(backend/shared): mapbox api integration ss-193

h4vefundieyoung:
Updates:
  - Added: http client.
  - Added: abstract api layer.
  - Added: map box api implementation.
  - Added: `/routes/construct`endpoint, that allows to build routes using POIs ids.
  - Added: debouncing logic for API requests, it's limited with 5 requests per minute.

https://github.com/user-attachments/assets/4672d9e3-60b2-4c47-8c6a-d7b039bcf042


apps/backend/src/libs/modules/config/base-config.module.ts
```diff
@@ -87,6 +87,14 @@ class BaseConfig implements Config {
 					format: Number,
 				},
 			},
+			MAPBOX_DIRECTIONS_API: {
```
liza-veis, line 90: let's just have everything related to mapbox under one group
```suggestion
			MAPBOX: {
```

apps/backend/src/libs/modules/map-box/libs/enums/map-box-paths.enum.ts
```diff
@@ -0,0 +1,5 @@
+const MapBoxAPIPaths = {
```
liza-veis, line 1: The enum name should be in singular form, the same for MapBoxAPIProfiles -> MapboxAPIProfile
```suggestion
const MapboxAPIPath = {
```

apps/backend/src/modules/points-of-interest/points-of-interest.repository.ts
```diff
@@ -59,6 +59,23 @@ class PointsOfInterestRepository implements Repository {
 		);
 	}
 
+	public async findAllById(ids: number[]): Promise<PointsOfInterestEntity[]> {
```
liza-veis, line 62: Take a look at how it's implemented here, we will pass IDs as a filter in options for findAll https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/104

apps/backend/src/modules/routes/routes.controller.ts
```diff
@@ -0,0 +1,205 @@
...
+
+		if (requestsAmount === MAX_REQUESTS_PER_MINUTE) {
+			throw new RoutesError({
+				message: RoutesExceptionMessage.LIMIT_REACHED,
+				status: HTTPCode.REQUEST_LIMIT,
+			});
+		}
+
+		this.debounceMap[id]++;
+	}
```
liza-veis, line 202: Maybe we can create a custom hook for rate limiting? to remove this logic from the controller + to make it reusable for routes. For now, we only need one implementation - use the user ID from the injected user, and let's pass requestLimitPerMin as parameter. We will only use it for routes where user should be present

apps/backend/src/libs/modules/map-box/map-box-directions-api.ts
```diff
@@ -0,0 +1,62 @@
...
+			alternatives: "false",
+			geometries: geometricsType,
+			overview: "simplified",
+			steps: "false",
+		});
+
+		const URLCoords = coordinates
+			.map((coordinate) => `${coordinate.join(",")};`)
+			.join("")
+			.replace(/.$/, "");
```
my-derkach, line 47: What do you think about that:
```suggestion
		const URLCoords = coordinates
			.map((coordinate) => coordinate.join(",")).join(";");
```

I think that the result must to be the same

apps/backend/src/modules/routes/routes.controller.ts
```diff
@@ -0,0 +1,205 @@
...
+	 */
+	public async constructRoute({
+		body: { pointsOfInterest },
+		user,
+	}: APIHandlerOptions<{ body: ConstructRouteRequestDto }>): Promise<
+		APIHandlerResponse<GetMapBoxRouteResponseDto>
+	> {
+		if (user) {
+			this.debounce(user.id);
+		}
```
liza-veis, line 171: We know by logic that the user will be present, so we don't need this check. In case something can't happen by logic, we will use a type cast instead of adding unnecessary checks

apps/backend/src/modules/routes/routes.service.ts
```diff
@@ -0,0 +1,50 @@
...
+
+	public constructor(
+		mapBoxDirectionsApi: MapBoxDirectionsApi,
+		pointOfInterestService: PointsOfInterestService,
+	) {
+		this.pointsOfInterestService = pointOfInterestService;
+		this.mapBoxDirectionApi = mapBoxDirectionsApi;
+	}
+
+	public async buildRoute(
```
liza-veis, line 25: We have a rule somewhere in the quality criteria to avoid name repetition in methods for entities, because their name is already obvious from the class name. So it should be either construct, or constructPath/contructDirections
```suggestion
	public async construct(
```

packages/shared/src/libs/modules/http/libs/enums/http-code.enum.ts
```diff
@@ -4,6 +4,7 @@ const HTTPCode = {
 	INTERNAL_SERVER_ERROR: 500,
 	NOT_FOUND: 404,
 	OK: 200,
+	REQUEST_LIMIT: 429,
```
liza-veis, line 7: let's use its standard name
```suggestion
	TOO_MANY_REQUESTS: 429,
```

packages/shared/src/modules/routes/libs/enums/route-api-path.ts
liza-veis, line 1: filename should end with .enum.ts

packages/shared/src/modules/routes/libs/types/get-map-box-route-response-dto.type.ts
```diff
@@ -0,0 +1,39 @@
+type Coordinate = [number, number];
```
liza-veis, line 1: All these types, except for maybe Coordinate, should be kept in the mapbox module on the backend; we don't want to expose any api internal implementation to the client. But I also think we won't need Coordinate for now either, if we need it, it's better to define use PointGeometry['coordinates'] which we already have

liza-veis (changes requested):
A few comments but great job overall 😎

apps/backend/src/libs/modules/map-box/map-box-directions-api.ts
```diff
@@ -0,0 +1,62 @@
...
+
+		const response = await this.load<GetMapBoxRouteResponseDto>(
+			this.getFullEndpoint(URL, {}),
+			{
+				contentType: ContentType.JSON,
+				method: "GET",
+			},
+		);
+
+		return await response.json();
```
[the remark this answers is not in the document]
liza-veis: I agree it's not necessary, but it improves error handling and debugging significantly by displaying the right context in the error stack. And also in our case, it's required by the linter to await all the async calls inside the function. Nevertheless, thanks for the comment 🐈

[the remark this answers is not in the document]
h4vefundieyoung: yea, lint will never give it a chance

packages/shared/src/modules/routes/libs/types/get-map-box-route-response-dto.type.ts
liza-veis: hm, maybe I missed this question on a daily. Even if interfaces fully matched, we would go from our common interface and then apply it to the mapbox module, so all the naming will start with something like RouteConstructResponseDto because we don't want to expose internal implementation to the frontend. Also, we only need to return fields that we need, and I see lots of unnecessary fields in those interfaces, some of them we may need to be remapped or renamed to follow application styleguides - e.g. everything should be in camel case

vyacheslavKhoroshev:
pull changes, my PR with POIs merged and we have a conflicts.

apps/backend/src/libs/hooks/debouncer/get-request-debouncer.hook.ts
```diff
@@ -0,0 +1,32 @@
+import { type Hook } from "../libs/types/types.js";
+import { DEBOUNCER_CONFIG } from "./libs/constants/constants.js";
+import { RequestLimitError } from "./libs/exceptions/exceptions.js";
+
+const getRequestDebouncer = (requestsLimitPerMinute: number): Hook => {
+	let debounceMap: Record<string, number> = {};
```
liza-veis, line 6: let's use map for this and make it more clear what we store here
```suggestion
	const userIdToRequestNumber: Map<UserAuthResponseDto['id'], number> = new Map();
```

apps/backend/src/libs/hooks/debouncer/get-request-debouncer.hook.ts
```diff
@@ -0,0 +1,32 @@
+import { type Hook } from "../libs/types/types.js";
+import { DEBOUNCER_CONFIG } from "./libs/constants/constants.js";
+import { RequestLimitError } from "./libs/exceptions/exceptions.js";
+
+const getRequestDebouncer = (requestsLimitPerMinute: number): Hook => {
```
liza-veis, line 5: let's rename it to something more clear e.g setRateLimit, applyRateLimit
```suggestion
const setRateLimit = (requestsLimitPerMinute: number): Hook => {
```

apps/backend/src/libs/hooks/debouncer/get-request-debouncer.hook.ts
```diff
@@ -0,0 +1,32 @@
...
+	let debounceMap: Record<string, number> = {};
+
+	return ({ user }, _, done): void => {
+		const id = user?.id as number;
```
liza-veis, line 9: We either need to assume that the user is present here and use type case for the user itself, or account for the case when not and throw an error. I think it's better to move on with the second approach here

apps/backend/src/libs/hooks/debouncer/get-request-debouncer.hook.ts
```diff
@@ -0,0 +1,32 @@
...
+			return;
+		}
+
+		const requestsAmount = debounceMap[id];
+
+		if (requestsAmount === requestsLimitPerMinute) {
+			throw new RequestLimitError();
+		}
+
+		debounceMap[id]++;
```
liza-veis, line 27: Let's avoid ++ as it's harder to read
```suggestion
		debounceMap[id] += 1;
```

apps/backend/src/libs/hooks/libs/types/hook.type.ts
```diff
@@ -0,0 +1,8 @@
+import { type ServerApplicationRouteParameters } from "~/libs/modules/server-application/server-application.js";
+
+type Hook = Exclude<
+	ServerApplicationRouteParameters["preHandlers"],
+	undefined
+>[0];
```
liza-veis, line 6: Let's define this type more explicitly 
```
type APIPreHandler = (options: APIHandlerOptions, done: () => void) => void;
```

liza-veis (changes requested):

apps/backend/src/libs/modules/mapbox/libs/types/get-mapbox-route-response-dto.type.ts
```diff
@@ -0,0 +1,39 @@
...
+	legs: MapboxLegs[];
+	weight: number;
+	weight_name: string;
+};
+
+type MapboxWaypoint = {
+	distance: number;
+	location: PointGeometry["coordinates"][];
+	name: string;
+};
```
liza-veis, line 37: Let's remove unnecessary fields and interfaces 
1. waypoints are deprecated, and we don't need them
2. We don't need legs for now, and not sure if we will. Also, we don't need weight
3. We only need to select the first recommendation per route. For this, we also need to pass `alternatives: false` to optimize the request
4. Eventually, before returning anything from this service, we need to remap it and return in the format and only the info we need 

```suggestion
type RouteContructResponseDto = { // if in shared
	route: {
	  distance: number;
	  duration: number;
	  geometry: LineStringGeometry
	}
	internalId: string; // uuid
};
```

As for geometry, I believe it should be of type
```
 {
    type: "LineString";  
    coordinates: PointGeometry["coordinates"][];
  };
```

If so, let's create a LineStringGeometry common type, similarly to PointGeometry

apps/backend/src/libs/modules/mapbox/libs/enums/mapbox-paths.enum.ts
liza-veis, line 1: file name should be mapbox-api-path.enum.ts and the same for other enums in this folder

apps/backend/src/libs/modules/mapbox/libs/enums/mapbox-geometrics-type.enum.ts
```diff
@@ -0,0 +1,7 @@
+const MapboxAPIGeometricsType = {
+	GEOJSON: "geojson",
+	POLYLINE: "polyline",
+	POLYLINE6: "polyline6",
```
liza-veis, line 4: let's remove options that are not used
```suggestion
```

apps/backend/src/libs/modules/config/base-config.module.ts
```diff
@@ -87,6 +87,20 @@ class BaseConfig implements Config {
...
+					doc: "Mapbox api access token",
+					env: "MAPBOX_ACCESS_TOKEN",
+					format: String,
+				},
+				MAPBOX_URL: {
+					default: null,
+					doc: "Mapbox URL",
+					env: "MAPBOX_URL",
+					format: String,
+				},
```
liza-veis, line 102: let's unitfy the naming in groups to avoid repetition
```suggestion
			MAPBOX: {
				ACCESS_TOKEN: {
					default: null,
					doc: "Mapbox api access token",
					env: "MAPBOX_ACCESS_TOKEN",
					format: String,
				},
				BASE_URL: {
					default: null,
					doc: "Mapbox base URL",
					env: "MAPBOX_BASE_URL",
					format: String,
				},
```

apps/backend/src/libs/modules/controller/base-controller.module.ts
```diff
@@ -24,13 +24,14 @@ class BaseController implements Controller {
...
 	): void {
-		const { handler, path } = options;
+		const { handler, path, preHandlers = [] } = options;
 		const fullPath = this.apiUrl + path;
 
 		this.routes.push({
 			...options,
 			handler: (request, reply) => this.mapHandler(handler, request, reply),
 			path: fullPath,
+			preHandlers,
```
liza-veis, line 34: It should have the same logic as the handler, we want to control the information we pass to it, not the whole req/reply


```suggestion
			preHandlers:  preHandlers.map((preHandler) => {
				return (request, _, done): void => {
					this.mapPreHandler(preHandler, request, done);
				};
			}),
```

apps/backend/src/libs/hooks/debouncer/set-rate-limit.hook.ts
```diff
@@ -0,0 +1,41 @@
...
+
+	return ({ user }, _, done): void => {
+		if (!user) {
+			throw new AuthError();
+		}
+
+		const { id } = user;
+		const requestsAmount = userIdToRequestNumber.get(id);
+
+		if (!requestsAmount) {
```
liza-veis, line 22: The naming should be uniform; if it's requestNumber in the map, then it should be requestNumber here too
```suggestion
		const requestNumber = userIdToRequestNumber.get(id);

		if (!requestNumber) {
```

apps/backend/src/libs/hooks/debouncer/set-rate-limit.hook.ts
```diff
@@ -0,0 +1,41 @@
...
+			userIdToRequestNumber.set(id, INCREMENT);
+
+			setTimeout(() => {
+				userIdToRequestNumber.set(id, 0);
+			}, DEBOUNCER_CONFIG.RESET_TIME);
+
+			return;
+		}
+
+		if (requestsAmount === requestsLimitPerMinute) {
```
liza-veis, line 32: it's always better to account for non standard concurrency cases
```suggestion
		if (requestsAmount >= requestsLimitPerMinute) {
```

apps/backend/src/libs/hooks/debouncer/set-rate-limit.hook.ts
```diff
@@ -0,0 +1,41 @@
+import { type UserAuthResponseDto } from "~/libs/types/types.js";
+import { AuthError } from "~/modules/auth/libs/exceptions/unauthorized.exception.js";
+
+import { type APIPreHandler } from "../libs/types/types.js";
+import { DEBOUNCER_CONFIG } from "./libs/constants/constants.js";
+import { RequestLimitError } from "./libs/exceptions/exceptions.js";
+
+const INCREMENT = 1;
```
liza-veis, line 8:
```suggestion
const REQUEST_NUMBER_INITIAL_VALUE = 0;
const REQUEST_NUMBER_INCREMENT_VALUE = 1;
```

apps/backend/src/libs/hooks/debouncer/libs/constants/debouncer-config.constant.ts
```diff
@@ -0,0 +1,5 @@
+const DEBOUNCER_CONFIG = {
```
liza-veis, line 1: Let's move TIME_UNIT constant from frontend to shared and use TIME_UNIT.MILLISECONDS_IN_SECOND * TIME_UNIT.SECONDS_IN_MINUTE

liza-veis (changes requested):

apps/backend/src/modules/routes/routes.service.test.ts
```diff
@@ -21,6 +24,18 @@ const FIRST_COORDINATE = 30.5234;
 const SECOND_COORDINATE = 50.4501;
 const FIRST_VISIT_ORDER = 0;
 const SECOND_VISIT_ORDER = 1;
+const NUMS_MAP = {
```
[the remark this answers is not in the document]
h4vefundieyoung: @liza-veis  it definetely worth it. such a pain for no reason. most values are 0 and 1, and they are obvious in 99% of cases, so these tons of constants just turning it into the mess. i may be wrong, but thats how i see it.

packages/shared/src/modules/routes/libs/validation-schemas/libs/enums/routes-validation-message.enum.ts
```diff
@@ -0,0 +1,9 @@
+import { ROUTES_VALIDATION_CONFIG } from "../constants/constants.js";
+
+const RoutesValidationMessage = {
+	INVALID_INPUT: "Points of interest should be a number typed array",
+	MAX_INPUT_LENGTH: `Input data shouldn't containt more than ${ROUTES_VALIDATION_CONFIG.REQUEST_MAX_POINTS_OF_INTEREST.toString()} items.`,
```
kalinkaaaa14, line 5:
```suggestion
	MAX_INPUT_LENGTH: `Input data shouldn't contain more than ${ROUTES_VALIDATION_CONFIG.REQUEST_MAX_POINTS_OF_INTEREST.toString()} items.`,
```

packages/shared/src/modules/routes/libs/validation-schemas/libs/constants/validation-config.ts
kalinkaaaa14, line 1: Filename shoud be routes-validation-config.constant.ts

packages/shared/src/modules/routes/libs/types/routes-response-construct-dto.type.ts
```diff
@@ -0,0 +1,12 @@
+import { type LineStringGeometry } from "../../../../libs/types/types.js";
+
+type RoutesResponseConstructDto = {
```
kalinkaaaa14, line 3:
```suggestion
type RouteConstructResponseDto = {
```

packages/shared/src/modules/routes/libs/types/routes-request-construct-dto.type.ts
```diff
@@ -0,0 +1,5 @@
+type RoutesRequestConstructDto = {
```
kalinkaaaa14, line 1:
```suggestion
type RouteConstructRequestDto = {
```

packages/shared/src/modules/routes/libs/enums/route-api-path.enum.ts
```diff
@@ -0,0 +1,5 @@
+const RouteApiPath = {
```
kalinkaaaa14, line 1: We have already routes module merged, but I see now that there wasn't api path enum created. Can you please add path that are already used in the routes controller to this enum, and reuse values from it in the controller?

packages/shared/src/libs/constants/time-unit.ts
kalinkaaaa14, line 1: the filename should be time-unit.constant.ts

packages/shared/src/libs/constants/constants.ts
```diff
@@ -1 +1,2 @@
 export { METERS_IN_KM } from "./meters-in-km.js";
```
kalinkaaaa14, line 1: fix filename here also please

apps/backend/src/libs/modules/mapbox/mapbox-directions-api.ts
```diff
@@ -0,0 +1,61 @@
...
+			path: MapboxAPIPath.DIRECTIONS,
+		});
+		this.accessToken = config.ENV.MAPBOX.ACCESS_TOKEN;
+	}
+
+	public async getRoute(
+		profile: ValueOf<typeof MapboxAPIProfile>,
+		coordinates: PointGeometry["coordinates"][],
+		geometricsType: ValueOf<typeof MapboxAPIGeometric>,
+	): Promise<MapboxConstructRouteResponseDto> {
```
kalinkaaaa14, line 35: When we have 3 or more args, it's better to pass them as an object, and not as list of params

apps/backend/src/libs/modules/mapbox/mapbox-directions-api.ts
```diff
@@ -0,0 +1,61 @@
...
+		const query = {
```
kalinkaaaa14, line 36: let's type this object also

apps/backend/src/libs/modules/mapbox/libs/enums/mapbox-profile.enum.ts
kalinkaaaa14, line 1: filename - mapbox-api-profile

apps/backend/src/libs/modules/mapbox/libs/enums/mapbox-path.enum.ts
kalinkaaaa14, line 1: filename - mapbox-api-path

apps/backend/src/libs/modules/mapbox/libs/enums/mapbox-geometric.enum.ts
kalinkaaaa14, line 1: filename - mapbox-api-geometric

apps/backend/src/modules/routes/routes.service.test.ts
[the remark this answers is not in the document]
liza-veis: yeah, in tests we usually use lots of magical numbers, so I agree that we need to adjust config to allow this

packages/shared/src/libs/constants/time-unit.ts
liza-veis, line 1: filename should be time-unit.constant.ts

packages/shared/src/modules/routes/libs/types/routes-request-construct-dto.type.ts
h4vefundieyoung: @kalinkaaaa14 I picked that naming pattern from the merged route branch, and routes used close to everywhere. should I rename all that stuff?

kalinkaaaa14: Hm, let's make it then RoutesConstructRequestDto & RoutesConstructResponseDto, thanks that you've noticed

packages/shared/src/modules/routes/libs/enums/route-api-path.enum.ts
kalinkaaaa14: For each controller we should have ApiPath enum, that was initial config. Now we have few modules with it created, and some of them don't. I think we need to add those enums where they are missing, but that could be done separately. I've just asked you to add those path that are used in routes module, just few values

apps/backend/src/libs/modules/mapbox/mapbox-directions-api.ts
kalinkaaaa14: Hm, I thought 3 is included into this rule) But if it's allowed, than ok, we can leave it as i) On the practice, for me it's always easier to understand what params the specific method has and how to use those specific methods when it has object with params.

packages/shared/src/modules/routes/libs/enums/route-api-path.enum.ts
h4vefundieyoung: I mean all values we have in the route module is root values, like "/" and "/:id", I've added "/construct" value for the endpoint I've implemented in the scope of current task. So the question is - should I add these root values as well?

apps/backend/src/libs/modules/controller/base-controller.module.ts
```diff
@@ -49,6 +53,24 @@ class BaseController implements Controller {
...
+		done: () => void,
+	) => void)[] {
+		return preHandlers.map((preHandler) => {
+			return (
+				{ body, params, query, user }: FastifyRequest,
+				_: FastifyReply,
+				done: () => void,
+			): void => {
+				preHandler({ body, params, query, user }, done);
+			};
```
liza-veis, line 70: We already have a mapper for the request, let's use it since it's returning the same interface
```suggestion
			return (
				request: FastifyRequest,
				_: FastifyReply,
				done: () => void,
			): void => {
			    const preHandlerOptions = this.mapRequest(request);
			    
				preHandler(preHandlerOptions, done);
			};
```

apps/backend/src/libs/modules/mapbox/libs/types/mapbox-construct-route-mapped-data.type.ts
```diff
@@ -0,0 +1,12 @@
+import { type LineStringGeometry } from "~/libs/types/types.js";
+
+type MapboxConstructRouteMappedData = {
```
liza-veis, line 3: let's call it by the app's common conventions
```suggestion
type MapboxConstructRouteResponseDto = {
```

 And as for internal to this module type `MapboxConstructRouteResponseDto` - we can name it like `MapboxConstructRouteApiResponseDto`, but its name is relatively less important than this one, because it can be exported and used in the backend in the future

liza-veis (changes requested):
Two small comments, everything else looks good 😎

liza-veis (approved):

packages/shared/src/modules/routes/libs/validation-schemas/libs/constants/routes-validation-config.constant.ts
```diff
@@ -0,0 +1,6 @@
+const ROUTES_VALIDATION_CONFIG = {
```
liza-veis, line 1: Even though I agree it should be in constants by logic, we are currently using enum for validation rules, so let's follow this convention and  move it to enums instead of libs, and it should be in routes/libs/enums/route-validation-rule.enum.ts alongside the route-validation-message.enum.ts, not under validation-schema folder
```suggestion
const RouteValidationRule = {
```

PR #239, opened 2025-08-06
test(backend): add auth api tests ss-236

### LINKED ISSUE #236: feat: Create auto tests for api
### What feature?

Create auto tests for API in authorization module

### Screenshots

_No response_

yuliiaaaaa:
Resolves #236


liza-veis:
@yuliiaaaaa I've updated the configs, a few points that are worth mentioning:

1. We will be using - `test(backend)` scope for the backend integration tests. So commit messages should look like this:
```
test(backend): add tests for auth api ss-236
```
Branch name template:
```
236-test-add-tests-for-auth-api
```
It should always start with 236-test as the type, and then be followed by a short description

2. It would be a bit resource-consuming for us to run tests on each pull-request update, since we are merging everything into the main right away, so I set it up to only run integration tests after something was merged into the main branch.
3. I set up the tests folder in a separate workspace, so if you need to run something from this folder or add additinal dependency, you need to do this by adding `-w tests` to the end, which means to execute something in the tests workspace
4. You need to copy .env.example, create a .env inside tests folder and replace the API_BASE_URL with the env server url as it was in config.
5. The command to run unit tests from the root:
```
npm run test:integration
```
6. In case you have any questions or need help with setup, config, linters, or would want to add e2e tests, don't hesitate to ask 🐱

tests/consts/validation-consts.ts
```diff
@@ -0,0 +1,11 @@
+const MAX_PASSWORD_LENGTH = 64;
+const MIN_PASSWORD_LENGTH = 6;
+
+const MAX_FIRSTNAME_LENGTH = 64;
+// TODO: uncomment when validation tests are implemented
+// const MIN_FIRSTNAME_LENGTH = 2;
+
+// const MAX_LASTNAME_LENGTH = 64;
+// const MIN_LASTNAME_LENGTH = 2;
```
lavandosya, line 9: do we still need this commented part?

lavandosya (approved):

tests/consts/validation-consts.ts
liza-veis: Removed

PR #243, opened 2025-08-06
fix(frontend): hide middleware errors in production ss-241

timosh-yura:
issue:
- getAuthUser action invokes justified rejection, which lead to handleError middleware catching it and displaying with toastNotifier

solution:
- if ENV is not production, handleErrorMiddleware is not included in frontend Store config


liza-veis:
As discussed on the call, we need to catch the exception and throw a custom one on the backend

timosh-yura:
passed the issue to Bohdan

PR #249, opened 2025-08-07
feat(backend/frontend): add logout functionality ss-182

my-derkach:
Add logout functionality on the backend and frontend pats:

**Frontend:**
* in auth api create method to call logout endpoint from backend
* create logout thunk that except of calling auth api, also drop token from storage
* divide header into two parts: authenticated header and unauthenticated header
* add dropdown menu component for authenticated header

https://github.com/user-attachments/assets/ec0cc241-9beb-4850-a79e-692531f7411e


apps/backend/src/libs/modules/token/token-blacklist.service.ts
h4vefundieyoung, line 1: if I'm getting it right, and that's token revocation service - I believe it should be implemented through the db.
this implementation a little bit unreliable. in case of server crash, production update etc data will be lost, so all of revoked tokens will become valid again.

apps/frontend/src/libs/components/header/unauthorized-header.tsx
```diff
@@ -0,0 +1,26 @@
...
+		label: string;
+		to: ValueOf<typeof AppRoute>;
+	}[];
+};
+
+const UnautharizedHeader = ({ actions }: Properties): JSX.Element => {
+	return (
+		<div className={styles["buttons"]}>
+			{actions.map(({ label, to }) => (
+				<Button key={to} label={label} to={to} type="button" />
```
h4vefundieyoung, line 20: all interactive elements should have inner element with description according to the accessibility quality criteria

apps/frontend/src/libs/components/user-dropdown/user-dropdown.tsx
```diff
@@ -0,0 +1,18 @@
...
+import styles from "./styles.module.css";
+
+type Properties = {
+	onLogout: () => void;
+};
+
+const UserDropdown = ({ onLogout }: Properties): JSX.Element => {
+	return (
+		<div className={styles["user-dropdown"]}>
+			<Button label="Log out" onClick={onLogout} />
```
h4vefundieyoung, line 13: same

apps/backend/src/libs/modules/token/token-blacklist.service.ts
my-derkach: Yeah, you are right. I do not implement it fully, because I want to ask coaches about that. In the ticket it was said that we need implement backend and frontend functionality, but I believe if we will have token expiration time to be quite small, then there is no need at the backend implementation at all

my-derkach:
@liza-veis I have a question regarding implementation of logout functionality on the backend part, when you will have time, please take a look on it. For now it is done with blacklist approach, but in current stage it is working in memory. I do not know for now exactly if need it, because for quite low expiration time for token, I think that frontend part would be enough, but as additional checks why not. What do you think about this approach? Or maybe here some others possible approaches to do that?

liza-veis:
> @liza-veis I have a question regarding implementation of logout functionality on the backend part, when you will have time, please take a look on it. For now it is done with blacklist approach, but in current stage it is working in memory. I do not know for now exactly if need it, because for quite low expiration time for token, I think that frontend part would be enough, but as additional checks why not. What do you think about this approach? Or maybe here some others possible approaches to do that?

It actually depends on the application requirements, for our application frontend part is enough for logout, and we don't need to blacklist tokens for the user

my-derkach:
> > @liza-veis I have a question regarding implementation of logout functionality on the backend part, when you will have time, please take a look on it. For now it is done with blacklist approach, but in current stage it is working in memory. I do not know for now exactly if need it, because for quite low expiration time for token, I think that frontend part would be enough, but as additional checks why not. What do you think about this approach? Or maybe here some others possible approaches to do that?
> 
> It actually depends on the application requirements, for our application frontend part is enough for logout, and we don't need to blacklist tokens for the user

So I can totally remove any backend implementation for it?

liza-veis:
yes, you approach is good, it's just our application doesn't require this logic

apps/frontend/src/libs/components/header/authenticated-header.tsx
```diff
@@ -0,0 +1,107 @@
...
+		};
+	}, [handleClickOutside, handleEscape]);
+
+	return (
+		<div className={styles["user-profile-wrapper"]} ref={dropdownReference}>
+			<div
+				className={styles["user-info"]}
+				onClick={handleUserClick}
+				onKeyDown={handleKeyDown}
+				role="button"
```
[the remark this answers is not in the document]
my-derkach: Actually it was done for accessibility like screen readers

apps/frontend/src/libs/components/header/authenticated-header.tsx
```diff
@@ -0,0 +1,74 @@
...
+				<div className={styles["name"]}>
+					{user.firstName} {user.lastName}
+				</div>
+				<span
+					className={combineClassNames(
+						styles["dropdown-arrow"],
+						isDropdownOpen && styles["open"],
+					)}
+				>
+					<Icon height={20} name="dropdownArrow" width={20} />
```
h4vefundieyoung, line 66: i may be wrong, but at some moment @liza-veis told that all icons in the design should have 24px size

liza-veis: Depends on the part of the design. What I meant is:
1. Currently, they are all 24px on design, so it's a preferable size for us unless it looks better with a different size 
2. If we export them, we need to export them in a frame that makes them squared in size + adds small padding, which helps with aligning the icons. And desirable to export them in the same size as other icons, but the used size may differ. In our case, I think sticking to the same size, 24, is a good idea

PR #257, opened 2025-08-07
feat(frontend): add styles to select

### LINKED ISSUE #255: feat: Increase gap between selects on dashboard page
### What feature?

Resize "Select component"

### Screenshots

Image

vldkorin:
Added styles to select clear indicator
hover:
image
Increased gap between selects and bottom of page
image

Closes #255

### DIFF
--- a/apps/frontend/src/libs/components/select/styles.module.css
+++ b/apps/frontend/src/libs/components/select/styles.module.css
@@ -81,6 +81,16 @@
 	background-color: var(--color-text-error);
 }
 
+.clear-indicator {
+	color: var(--color-text-strong);
+	cursor: pointer;
+}
+
+.clear-indicator:hover {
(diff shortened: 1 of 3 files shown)


apps/frontend/src/libs/components/select/styles.module.css
```diff
@@ -81,6 +81,16 @@
...
 
+.clear-indicator {
+	color: var(--color-text-strong);
+	cursor: pointer;
+}
+
+.clear-indicator:hover {
+	color: var(--color-base);
+	background-color: var(--color-text-error);
+}
```
liza-veis, line 92: Red looks too bright for me in this case 😁 Maybe we can use background-hover for background color and leave the font color as it is, also please add some transition to it

liza-veis (approved):

PR #262, opened 2025-08-07
fix(frontend/backend): handle jwt/auth errors appearance ss-248

h4vefundieyoung:
Updates:
  -  Added try catch statement in auth plugin for jwt validation
  - Excluded toast notifications for auth errors on the client

### DIFF
--- a/apps/frontend/src/libs/modules/http/libs/exceptions/auth.exception.ts
+++ b/apps/frontend/src/libs/modules/http/libs/exceptions/auth.exception.ts
@@ -1,0 +1,37 @@
+import { type APIErrorType } from "~/libs/enums/enums.js";
+import { type HTTPCode } from "~/libs/modules/http/http.js";
+import {
+	type APIValidationErrorDetail,
+	type ValueOf,
+} from "~/libs/types/types.js";
+
+import { HTTPError } from "./http-error.exception.js";
+
+type Constructor = {
+	cause?: unknown;
+	details: APIValidationErrorDetail[];
+	message: string;
+	name?: string;
+	status: ValueOf<typeof HTTPCode>;
+	type: ValueOf<typeof APIErrorType>;
+};
+
+const AUTH_ERROR_DEFAULT_NAME = "Unauthorized";
+
+class AuthError extends HTTPError {
+	public name: string;
+
+	public constructor({
+		cause,
+		details,
+		message,
+		name = AUTH_ERROR_DEFAULT_NAME,
+		status,
+		type,
+	}: Constructor) {
+		super({ cause, details, message, status, type });
+		this.name = name;
+	}
+}
+
+export { AUTH_ERROR_DEFAULT_NAME, AuthError };
--- a/apps/frontend/src/libs/modules/store/middlewares/handle-error.middleware.ts
+++ b/apps/frontend/src/libs/modules/store/middlewares/handle-error.middleware.ts
@@ -1,12 +1,17 @@
 import { isRejected, type Middleware } from "@reduxjs/toolkit";
 
 import { CommonExceptionMessage } from "~/libs/enums/enums.js";
+import { AUTH_ERROR_DEFAULT_NAME } from "~/libs/modules/http/http.js";
 import { type ExtraArguments } from "~/libs/modules/store/libs/types/types.js";
 
 const handleError = ({ toastNotifier }: ExtraArguments): Middleware => {
 	return () => {
 		return (next) => (action) => {
 			if (isRejected(action)) {
+				if (action.error.name === AUTH_ERROR_DEFAULT_NAME) {
+					return;
+				}
+
 				toastNotifier.showError(
 					action.error.message ??
 						CommonExceptionMessage.COMMON_EXCEPTION_MESSAGE,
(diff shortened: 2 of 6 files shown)


apps/frontend/src/libs/modules/store/middlewares/handle-error.middleware.ts
```diff
@@ -1,12 +1,17 @@
...
 import { type ExtraArguments } from "~/libs/modules/store/libs/types/types.js";
 
 const handleError = ({ toastNotifier }: ExtraArguments): Middleware => {
 	return () => {
 		return (next) => (action) => {
 			if (isRejected(action)) {
+				if (action.error.name === AUTH_ERROR_DEFAULT_NAME) {
+					return;
+				}
+
```
kalinkaaaa14, line 14: Currently it will break the login for login page. For example, if user send invalid password and sent the request - he will never see the error from the server, and that's invalid behavior. 

So the logic should work like this - if we receive 401 on any of the pages - redirect to sign in, no toast notification is shown for the user.

If user is on sign in page and receives 401 - errors should be visible to him

kalinkaaaa14 (changes requested):

apps/frontend/src/libs/modules/store/middlewares/handle-error.middleware.ts
kalinkaaaa14: I think it should be done in the scope of the current ticket. And probably @liza-veis meant it from the beginning, maybe some miscommunication happened

liza-veis: Yes, we want to hide all 401 responses, except for the case of the login action. It wasn't miscommunication, I just forgot about this case, and that here we need to return an error message here. @kalinkaaaa14 thanks for pointing out. We definitely need to do this in this ticket, because it would mean that we will break existing functionality, which we obviously don't want, even if it will look slightly better for the demo 😅

liza-veis: It shouldn't be hard, though, we can just import the sign in action here and exclude it in the check if it matches. @h4vefundieyoung, also, please mention this case in the ticket, so that qa were aware of this case and could test it properly

apps/frontend/src/libs/modules/http/libs/exceptions/auth.exception.ts
```diff
@@ -0,0 +1,37 @@
...
+import { HTTPError } from "./http-error.exception.js";
+
+type Constructor = {
+	cause?: unknown;
+	details: APIValidationErrorDetail[];
+	message: string;
+	name?: string;
+	status: ValueOf<typeof HTTPCode>;
+	type: ValueOf<typeof APIErrorType>;
+};
```
liza-veis, line 17: We actually don't need to pass the name to the auth error, name is something internal and specific for this class.
```suggestion
```

 Instead, let's create a common enum - ExceptionName

```
const ExceptionName = {
  AUTH_ERROR = 'AuthError'
}
```

liza-veis (changes requested):

apps/frontend/src/libs/modules/store/middlewares/handle-error.middleware.ts
kalinkaaaa14: @liza-veis @h4vefundieyoung  Also was thinking about this case on login page, and actually we can avoid it at all, because what is the reason to return 401 on sign in request? If credentials are invalid - we should sent 400, it's a bad request with incorrect payload. So, the easiest way i think will be just to change error code from 401 to 400 when we are sending INVALID_CREDENTIALS error

h4vefundieyoung: @kalinkaaaa14 yeah, agree, good solution

liza-veis: I think 401 is more appropriate semantically here and more widely used for such cases:
- 400 - Either the request is malformed, e.g., we expect a string, but pass a number, or the structure of the request is invalid. 
- 422 - The request is in correct format, but can't be processed, e.g., because of internal logic rules.
- 401 - Indicates authentication is required and either missing or invalid, which is our case

h4vefundieyoung: @liza-veis well, makes sense then. i've implemented it with action type check.

apps/frontend/src/libs/modules/http/libs/exceptions/auth.exception.ts
```diff
@@ -0,0 +1,29 @@
...
+
+class AuthError extends HTTPError {
+	public static readonly name = "Unauthorized";
+	public name: string;
```
liza-veis, line 21: having it as a static and internal property is a bit confusing, let's only keep it as a property and use ExceptionName enum for matching. Also, instead of "Unauthorized", it's better to either name it 'AuthError' - to match the error, or to rename the error to Unauthorized error. I like the first option more
```suggestion
	public name: string;
```

PR #265, opened 2025-08-07
feat(frontend): implement password visibility toggler ss-184

h4vefundieyoung:
Changes:
  - Added input button to show/hide credentials
  

https://github.com/user-attachments/assets/cce5640e-e39e-4c81-8ec8-b267062e2a7d


apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -30,32 +31,60 @@ const Input = <T extends FieldValues>({
...
+					type={typeState}
+					value={field.value}
+				/>
+				{hasError && (
+					<span className={styles["error"]}>
+						<Icon height={24} name="error" width={24} />
+						{error as string}
+					</span>
+				)}
+				{isPasswordButtonShown && field.value && (
```
lavandosya, line 72: already have this check before
```suggestion
				{isPasswordButtonShown && (
```

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -30,32 +31,60 @@ const Input = <T extends FieldValues>({
...
 	type = "text",
 }: Properties<T>): React.JSX.Element => {
+	const [typeState, setTypeState] = useState<Properties<never>["type"]>(type);
 	const { field } = useFormController({ control, name });
 
 	const error = errors[name]?.message;
 	const hasError = Boolean(error);
+	const isPasswordButtonShown = type === "password" && field.value;
+
+	const onPassAppearanceToggle = useCallback((): void => {
```
lavandosya, line 41:
```suggestion
	const handlePasswordAppearanceToggle = useCallback((): void => {
```

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -30,32 +31,60 @@ const Input = <T extends FieldValues>({
...
+		const state: "password" | "text" =
+			typeState === "password" ? "text" : "password";
+
+		setTypeState(state);
+	}, [typeState, setTypeState]);
+	const toggleButtonIcon: "eye" | "eyeOff" =
+		typeState === "text" ? "eyeOff" : "eye";
```
lavandosya, line 48: Consider moving the password visibility toggle logic and button into a separate component. This will help keep the Input component cleaner and more focused, reducing its responsibility for show/hide password logic. Separating this logic makes the code easier to maintain and improves readability

lavandosya, line 48:
```suggestion
	const typeStateToPasswordIcon: "eye" | "eyeOff" = {
	  'password': 'eye',
	  'text': 'eyeOff'
	}
```

lavandosya (changes requested):

lavandosya:
We could make a small UI improvement, when clicking the "show/hide password" button, the highlighted border on the input field shouldn't be removed

apps/frontend/src/libs/components/input/libs/components/toggle-vision-button/toggle-vision-button.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+	return (
+		<button
+			className={styles["password-toggle-button"]}
+			onMouseDown={handleClick}
+			type="button"
+		>
+			<span className="visually-hidden">toggle password appearance</span>
+			<Icon
+				className={styles["password-toggle-button-icon"] || ""}
```
kalinkaaaa14, line 35:
```suggestion
				className={styles["password-toggle-button-icon"]}
```

apps/frontend/src/libs/components/input/libs/components/toggle-vision-button/toggle-vision-button.tsx
```diff
@@ -0,0 +1,44 @@
...
+type Properties = {
+	onAppearanceToggle: () => void;
+	type: "password" | "text";
+};
+
+const ToggleVisionButton = ({
+	onAppearanceToggle,
+	type,
+}: Properties): React.JSX.Element => {
+	const typeStateToPasswordIcon: "eye" | "eyeOff" =
```
kalinkaaaa14, line 16: typeStateToPasswordIcon sounds more like a map for me 🤔 

```suggestion
	const iconName: "eye" | "eyeOff" =
```

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -30,32 +32,51 @@ const Input = <T extends FieldValues>({
...
+	const isPasswordButtonShown =
+		(type === "password" && typeState === "password") ||
+		(typeState === "text" && field.value);
+
+	const handlePasswordAppearanceToggle = useCallback((): void => {
+		const state: "password" | "text" =
+			typeState === "password" ? "text" : "password";
+
+		setTypeState(state);
+	}, [typeState, setTypeState]);
```
kalinkaaaa14, line 49: I was thinking about this approach, and I'd prefer not to manipulate with the type within input component like this, and to have logic that relates to password field separately. What do you think about creating new PasswordInput, which will contain logic related to password inputs, and will reuse this Input inside? And also part with ToggleVision will be placed there

apps/frontend/src/libs/components/input/libs/components/toggle-vision-button/toggle-vision-button.tsx
h4vefundieyoung: it comes as styles | undefined from there, i can't pass it to the custom component like this

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -35,27 +41,55 @@ const Input = <T extends FieldValues>({
...
+				)}
+				{icon && (
+					<button
+						className={styles["input-button"]}
+						onMouseDown={handleIconClick}
+						type="button"
+					>
+						<span className="visually-hidden">input icon-button</span>
+						<Icon
+							className={styles["input-button-icon"] || ""}
```
kalinkaaaa14, line 85: Now we are moving with such an approach that we are not passing className styles from the outside of the component, can you make a wrapper for this icon if you need to apply some specific styles?

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -35,27 +41,55 @@ const Input = <T extends FieldValues>({
...
+						<Icon height={24} name="error" width={24} />
+						{error as string}
+					</span>
+				)}
+				{icon && (
+					<button
+						className={styles["input-button"]}
+						onMouseDown={handleIconClick}
+						type="button"
+					>
```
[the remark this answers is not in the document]
h4vefundieyoung: @liza-veis is there really a reason to have two of these containers, span/button i mean. can't we just disable the button in case it won't be interactable?

apps/frontend/src/libs/components/input/input.tsx
h4vefundieyoung: > Agree. Our Icon component is adjusted to inherit font color. To make this work, when we import icons, we need to set the property that gives a color to the icon to `currentColor` - css placeholder. Depending on the icon, it could be fill or stroke color properties in SVG. Also, we need to delete stroke-opacity if present. Then, to apply the necessary color to the Icon, we need to wrap it in a container outside, e.g., a div, and set a color property to it.
> 
> It's also important to note that we only do these steps for common icons, which we want to use similarly to a font (Icon component). For SVG images (not icons), e.g., logo, we leave them as it is.

i've never heard of it, sounds great, thanks

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -34,28 +42,59 @@ const Input = <T extends FieldValues>({
...
+			iconRight?.onClick?.();
+		},
+		[iconRight],
+	);
+
+	const iconComponent = iconRight && (
+		<span className={styles["input-button-icon"]}>
+			<Icon height={24} name={iconRight.name} width={24} />
+		</span>
+	);
```
liza-veis, line 59: I would define the icon component separately for interactive non non-interactive elements. But if you want to keep it this way, wrap it in useMemo and rename it to IconRightComponent - starting with a capital letter. + the class name should be logical and not related to buttons

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -34,28 +42,59 @@ const Input = <T extends FieldValues>({
...
+				)}
+				{iconRight && !isIconRightInteractive && iconComponent}
+				{iconRight && isIconRightInteractive && (
+					<button
+						className={styles["icon-button"]}
+						onMouseDown={handleIconClick}
+						type="button"
+					>
+						<span className="visually-hidden">
+							{iconRight.label || "input icon-button"}
```
liza-veis, line 92:
```suggestion
							{iconRight.label ?? "input icon-button"}
```

apps/frontend/src/libs/components/input/styles.module.css
```diff
@@ -22,23 +22,24 @@
...
-.input.input-error {
-	outline: none;
-	border: 2px solid var(--color-stroke-error-strong);
-}
-
-.input-error,
-.input-error:hover,
-.input-error:focus {
-	background-color: var(--color-fill-error-weak);
+.input-iconed {
```
liza-veis, line 41: Considering how we name other states, it should be input-icon-right, or at least input-with-icon-right; all nested components should not repeat input in their class names, unless it's a part of their name. We already know they are related to input by the component name. The same approach to using modules in the app overall - do not repeat the context name when it's clear from the context

apps/frontend/src/libs/components/input/styles.module.css
```diff
@@ -66,10 +78,33 @@
 
 @media (width <= 768px) {
 	.input {
-		padding: 8px 10px;
+		padding: 8px 50px 8px 10px;
```
liza-veis, line 81: padding 50px should only be applied for input with icon right

liza-veis (changes requested):

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -34,28 +45,40 @@ const Input = <T extends FieldValues>({
...
-				<span className={styles["error"]}>
-					<img alt="error-icon" src={errorIcon} />
-					{error as string}
+				<span className={styles["icon-right"]}>
+					{iconRight &&
+						(isInteractiveIcon ? (
+							<InteractiveIcon {...iconRight} />
+						) : (
+							<NonInteractiveIcon {...iconRight} />
+						))}
```
lavandosya, line 79:
```suggestion
		{iconRight &&
				<span className={styles["icon-right"]}>
						(isInteractiveIcon ? (
							<InteractiveIcon {...iconRight} />
						) : (
							<NonInteractiveIcon {...iconRight} />
						))}
```

apps/backend/src/db/migrations/20250813134000_add_users_profile_visibility.ts
lavandosya, line 1: what the purpose of this migration?

h4vefundieyoung: @lavandosya most likely accidently applied stash, i'll remove it

liza-veis (approved):

lavandosya (approved):

kalinkaaaa14 (approved):

PR #266, opened 2025-08-08
feat(backend): add an ability to search POIs by name ss-246

nazartsya:

- added an ability to search POIs by name
- the search should be case-insensitive


apps/backend/src/modules/points-of-interest/points-of-interest.controller.ts
```diff
@@ -210,7 +210,7 @@ class PointsOfInterestController extends BaseController {
 	 *     description: |
 	 *       Get all points of interest or filter them by location.
 	 *
-	 *       **Without query parameters**: Returns all points of interest
+	 *       **With query parameters**: Returns all points of interest. Supports optional search by name.
```
kalinkaaaa14, line 213: I think you need to leave the description as it was, because initial description was correct - without query parameters - Returns all points of interest

Make a separate description about searching by name

kalinkaaaa14 (approved):

PR #283, opened 2025-08-09
feat(frontend/backend): routes detail page ss-268

timosh-yura:
Added:

- routes reducer and getRoute action
- routes page at "/app/routes/{:id}"
- routes get api to whiteRoutes

Changed backend:

- routes description row increased varchar length to 2k
 
image


apps/frontend/src/pages/route/libs/components/images-section/styles.module.css
```diff
@@ -0,0 +1,43 @@
+.container {
+	display: flex;
+	gap: 16px;
+}
+
+.container img {
```
h4vefundieyoung, line 6: can't be used according to the quality criteria

apps/frontend/src/pages/route/libs/components/images-section/styles.module.css
```diff
@@ -0,0 +1,43 @@
...
+	border-radius: 8px;
+}
+
+.primary-description-wrapper {
+	display: flex;
+	flex-direction: column;
+}
+
+.primary-image {
+	height: 416px;
```
h4vefundieyoung, line 18: are you sure that we should use that fixed height?

apps/frontend/src/pages/route/libs/components/images-section/styles.module.css
```diff
@@ -0,0 +1,43 @@
...
+
+.secondary-images-wrapper {
+	display: flex;
+	flex-direction: column;
+	gap: 16px;
+	min-width: 359px;
+}
+
+.secondary-image {
+	height: 200px;
```
h4vefundieyoung, line 29: same

timosh-yura: Last time Lisa covered this topic, it was suggested that fixed height may be ok for the image height

timosh-yura: I`m not sure how to style it otherwise

apps/frontend/src/pages/route/libs/components/images-section/styles.module.css
```diff
@@ -0,0 +1,44 @@
+.container {
```
liza-veis, line 1: Let's use a grid instead. In short, we need 1 column which takes two rows and 2 columns where 1 image takes one row and the other one another row. For mobile screens, let's just keep them one row and add overflow auto, to make it scrollable to the right, here is an example
https://codepen.io/Tkrish/pen/xxZpwMK

liza-veis (changes requested):

apps/frontend/src/pages/route/libs/components/images-section/styles.module.css
liza-veis: i also think that here we can use max height instead with height: 100%

apps/frontend/src/pages/route/libs/components/images-section/styles.module.css
timosh-yura: Please describe more detailed

liza-veis: I mean, currently we have 2 rows with images. On smaller screens, it won't fit well, so we need to somehow adjust it. One way is to make all of them 1 row (the same size) and make them horizontally scrollable. At least in my imagination, it seems like it will look good here, but if you have other suggestions for mobile view, you can implement them. Also, let's make a common container for the page on bigger screens smaller, to maintain the same ratio between the large image and the smaller ones

timosh-yura:
Added:

- whiteRoutes path behaves like a regex

Chagned

- description not on the same column as a primary image, is this ok?

-  max-width for big screens

image

- Added horizontal scroll for mobile screens

image

apps/frontend/src/pages/route/route-details.tsx
```diff
@@ -0,0 +1,60 @@
...
+					<p className={styles["description"]}>{description}</p>
+				</div>
+			</main>
+		);
+	}
+
+	return <Loader />;
+};
+
+export { RouteDetails as Route };
```
kalinkaaaa14, line 60:
```suggestion
export { RouteDetails };
```

packages/shared/src/modules/routes/libs/types/routes-response-dto.type.ts
```diff
@@ -1,4 +1,13 @@
...
+type RouteGetAllItemResponseDto = {
+	id: number;
+	name: string;
+	pois: {
+		id: number;
+		visitOrder: number;
+	}[];
+}[];
+
+type RouteGetByIdResponseDto = {
```
kalinkaaaa14, line 10: these types should be placed in different files, and do not forget about filenames

package.json
```diff
@@ -69,5 +69,8 @@
 	"simple-git-hooks": {
 		"pre-commit": "npm run lint:types && npm run git:hooks:precommit",
 		"commit-msg": "npm run git:hooks:commit"
+	},
+	"dependencies": {
+		"path-to-regexp": "8.2.0"
```
kalinkaaaa14, line 74: Seems this lib is used only on backend, so install it not in the root package please, but on the backend

apps/frontend/src/modules/routes/routes-api.ts
```diff
@@ -0,0 +1,37 @@
...
+class RoutesApi extends BaseHTTPApi {
+	public constructor({ baseUrl, http, storage }: Constructor) {
+		super({ baseUrl, http, path: APIPath.ROUTES, storage });
+	}
+
+	public async getRouteById(
+		id: number,
+	): Promise<APIResponse<RouteGetByIdResponseDto>> {
+		const response = await this.load<APIResponse<RouteGetByIdResponseDto>>(
+			this.getFullEndpoint(`${RoutesApiPath.ROOT}${id.toString()}`, {}),
```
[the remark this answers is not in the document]
timosh-yura: indeed
  26:49  error  Invalid type "number" of template literal expression  @typescript-eslint/restrict-template-expressions

timosh-yura:
Updated, "routes" store would be in the same module or I should also rename every filename in "routes" dir to singular?

https://github.com/user-attachments/assets/65804f2e-4f54-423e-957a-9cc5fc6f6edb

apps/frontend/src/pages/route-description/libs/components/images-section/images-section.tsx
```diff
@@ -0,0 +1,34 @@
...
+					alt="Secondary point of interest of the route"
+					className={styles["secondary-image"]}
+					src={image2}
+				/>
+				<img
+					alt="Secondary point of interest of the route"
+					className={styles["secondary-image"]}
+					src={image3}
+				/>
+			</div>
```
liza-veis, line 29: We don't need all the wrappers around images; this can be done easily with a grid. I tried to play a bit with it and was able to achieve this result. Pay attention to how responsiveness works here, and obviously, we need to adjust styles according to conventions. This was just a draft example

https://codepen.io/liza-veis/pen/bNVYEJz

Also let's rename this component to ImageGallery

apps/frontend/src/pages/route-description/route-details.tsx
```diff
@@ -0,0 +1,60 @@
...
+				<div className={styles["container"]}>
+					<label className={styles["label"]}>{name}</label>
+					<ImagesSection />
+					<p className={styles["description"]}>{description}</p>
+				</div>
+			</main>
+		);
+	}
+
+	return <Loader />;
```
liza-veis, line 57: let's return the component the last one, and loader if status is pending

apps/frontend/src/pages/route-description/styles.module.css
```diff
@@ -0,0 +1,34 @@
+.container {
```
liza-veis, line 1: Let's leave a to-do comment here that these styles should be moved to the common layout

apps/frontend/src/pages/route-description/route-details.tsx
```diff
@@ -0,0 +1,60 @@
...
+				<Header
+					actions={[{ label: "Sign in", to: AppRoute.SIGN_IN }]}
+					user={authenticatedUser}
+				/>
+				<div className={styles["container"]}>
+					<label className={styles["label"]}>{name}</label>
+					<ImagesSection />
+					<p className={styles["description"]}>{description}</p>
+				</div>
+			</main>
```
liza-veis, line 53: Let's get rid of this unnecessary wrapper. We will move this to the common layout later. Also header should be h1, not label. And FYI, each page should have an H1 title, even if it's hidden, it should be present in the layout
```suggestion
               <>
				<Header
					actions={[{ label: "Sign in", to: AppRoute.SIGN_IN }]}
					user={authenticatedUser}
				/>
			<main className={styles["container"]}>
					<h1 className={styles["title"]}>{name}</label>
					<ImagesSection />
					<p className={styles["description"]}>{description}</p>
			</main>
			</>
```

packages/shared/src/modules/routes/libs/types/routes-response-get-all-dto.ts
liza-veis, line 1: fileName should end with .type.ts

liza-veis (changes requested):
Overall looks good, here are a few comments 🐱

apps/backend/src/db/migrations/20250810152620_alter_routes_description_length.ts
```diff
@@ -0,0 +1,26 @@
+import { type Knex } from "knex";
+
+const DESCRIPTION_LENGTH = 2000;
+const DEFAULT_DESCRIPTION_LENGTH = 255;
+
+async function down(knex: Knex): Promise<void> {
+	await knex("routes")
```
kalinkaaaa14, line 7: run these calls in a transaction please

timosh-yura:
what`s the convo on filenames?
not sure what`s wrong with "routes-response-get-all-dto.type.ts", changed to "routes-response-get-all-item-dto.type.ts"

timosh-yura:
Do I need to change something for the merge?

liza-veis (changes requested):

apps/backend/src/modules/routes/routes.controller.ts
```diff
@@ -7,14 +7,15 @@ import {
 } from "~/libs/modules/controller/controller.js";
 import { type Logger } from "~/libs/modules/logger/logger.js";
 
-import { RouteApiPath } from "./libs/enums/enums.js";
+import { RoutesApiPath } from "./libs/enums/enums.js";
```
h4vefundieyoung, line 10: I'm not really sure, these naming stuff getting confusing, but I believe we should name enums in singular form. So it should be RouteApiPath

apps/backend/src/modules/routes/routes.controller.ts
```diff
@@ -428,7 +429,7 @@ class RoutesController extends BaseController {
 
 	public async findById(
 		options: APIHandlerOptions<{ params: { id: string } }>,
-	): Promise<APIHandlerResponse<RoutesResponseDto>> {
+	): Promise<APIHandlerResponse<RouteGetByIdResponseDto>> {
```
h4vefundieyoung, line 432: I'm not really sure, the naming stuff is getting confusing, but it's routes everywhere, shouldn't we follow this pattern?

packages/shared/src/modules/routes/libs/types/routes-response-get-all-item-dto.type.ts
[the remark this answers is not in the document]
h4vefundieyoung: @kalinkaaaa14 so we're going to use route and not routes as convention, right?

apps/backend/src/modules/routes/routes.controller.ts
timosh-yura: I originally picked up naming pattern from points-of-interest (plural)
should i rename every file to singular?

kalinkaaaa14: @timosh-yura no, fix only places that are highlighted in this pr, we will fix the naming through the project in separate ticket

packages/shared/src/modules/routes/libs/types/route-update-request.dto.ts
kalinkaaaa14, line 1: route-patch-request-dto.type.ts

packages/shared/src/modules/routes/libs/types/routes-get-id-response-dto.type.ts
kalinkaaaa14, line 1: route-get-by-id-response-dto.type.ts

packages/shared/src/modules/routes/libs/types/routes-request-dto.type.ts
kalinkaaaa14, line 1: route-create-request-dto.type.ts

timosh-yura:
*fixed

Issue #188, opened 2025-08-01
feat: Search closest points to authenticated user


kalinkaaaa14:
- [x] Modify existing findAll method in POI module, that it will receive current user geolocation, and regarding this, will return the closest POI saved in our system. The radius of searching should be also thrown in the request query params.
- [x] If geolocation is not thrown - return all pois that are stored in the DB

vyacheslavKhoroshev:
https://github.com/user-attachments/assets/ce8462e9-207a-4d18-8174-0425d825fc27

vyacheslavKhoroshev:
- branch 188-poi-search based on 112-feat-poi-postgis;
- all commits with wrong issue‑number (118 instead 188) ;)

PR #299, opened 2025-08-11
feat(frontend): create poi frontend flow ss-244

### LINKED ISSUE #244: feat: Add an ability to create poi on frontend
### What feature?

- [ ] Create common modal component (use dialog tag)
- [ ] Implement "Create new POI" modal
  - For now, only add name and description and hard-code coordinates
- [ ] Add create POI button somewhere on dashboard page to test it until #189 is ready
- [ ] On success, the user should see a success toast message
- [ ] On error, the modal should remain open

### Screenshots

Image

vldkorin:
Added components:
- modal
- textarea
- create-poi-modal
Applied fixes to backend/shared:
- added description as optional field
- added description to validation schemas, dtos
Closes #244 

https://github.com/user-attachments/assets/5be1085e-3a7c-48cd-add2-be02e6c15b68


vldkorin:
> ### Create new POI
> ```
> 		<form
> 			className={styles["form"]}
> 			onSubmit={handleSubmit(handleFormSubmit)}
> 		>
> 			<div className={styles["field"]}>
> 				<Input
> 					autocomplete="name"
> 					control={control}
> 					errors={errors}
> 					label="Name"
> 					name="name"
> 					type="text"
> 				/>
> 			</div>
> 			<div className={styles["field"]}>
> 				<TextArea
> 					control={control}
> 					errors={errors}
> 					label="Description (optional)"
> 					name="description"
> 				/>
> 			</div>
> 
> 			<div className={styles["footer"]}>
> 				<div className="button">
> 					<Button label="Create" type="submit" />
> 					<div />
> 				</div>
> 			</div>
> 		</form>
> 	</Modal>
> );
> ```
> 
> image
> if we are going to use the same form for editing POI but already with filled in data, then maybe it is better to put your form in a separate component
> 
> PoiForm title="Create or Edit POI" control={control} and so on
> 
> right now it's just my guess because I don't know exactly how the editing should happen

I was thinking about that and i even try to do that by adding mode(create/update) property to poi modal, but my task was to implement only create modal, so i'll wait for coaches answ

kalinkaaaa14:
Yes, I think we can do it separately

apps/backend/src/db/migrations/20250801152138_add_test_poi_data.ts
```diff
@@ -4,38 +4,44 @@ const TABLE_NAME = "points_of_interest";
 
 const createTestPoiData = (
 	knex: Knex,
-): { location: Knex.Raw; name: string }[] => [
+): { description: string; location: Knex.Raw; name: string }[] => [
```
kalinkaaaa14, line 7: We can't change migrations that are already on the environment. 

How it works - migration is on env - it's already applied on the db, you open new PR with updating migration that's already applied - merge it to the env - nothing changes on the db, because it was previously applied.

apps/backend/src/db/migrations/20250801152137_add_description_to_points_of_interest.ts
```diff
@@ -0,0 +1,17 @@
+import { type Knex } from "knex";
+
+const TABLE_NAME = "points_of_interest";
+
+function down(knex: Knex): Promise<void> {
+	return knex.schema.alterTable(TABLE_NAME, (table) => {
+		table.dropColumn("description");
```
kalinkaaaa14, line 7: Description is optional, it should be nullable column

apps/backend/src/modules/points-of-interest/points-of-interest.controller.ts
```diff
@@ -49,13 +49,18 @@ import {
  *         name:
  *           type: string
  *           example: "Central Park"
+ *         description:
+ *           type: string
```
kalinkaaaa14, line 53:
```suggestion
 *           type: string | null
```

apps/frontend/src/pages/dashboard/dashboard.tsx
```diff
@@ -31,12 +55,40 @@ const Dashboard = (): React.JSX.Element => {
...
+				location: {
+					coordinates: payload.location.coordinates,
+					type: LocationType.POINT,
+				},
+				name: payload.name,
+			};
+
+			const result = await dispatch(
+				poiActions.create(dto as unknown as PointsOfInterestRequestDto),
+			);
```
kalinkaaaa14, line 82:
```suggestion
			const dto: PointsOfInterestRequestDto = {
				description: payload.description,
				name: payload.name,
			};

			const result = await dispatch(
				poiActions.create(dto),
			);
```

apps/frontend/src/modules/points-of-interest/slices/points-of-interest.slice.ts
```diff
@@ -0,0 +1,39 @@
...
+	dataStatus: DataStatus.IDLE,
+};
+
+const { actions, name, reducer } = createSlice({
+	extraReducers(builder) {
+		builder.addCase(create.pending, (state) => {
+			state.dataStatus = DataStatus.PENDING;
+		});
+		builder.addCase(create.fulfilled, (state, action) => {
+			state.data = Array.isArray(action.payload.data)
```
kalinkaaaa14, line 25: Is there a case when it will be an array?

packages/shared/src/modules/points-of-interest/libs/enums/points-of-interest-api-path.enum.ts
```diff
@@ -0,0 +1,5 @@
+const PointsOfInterestApiPath = {
+	CREATE: "",
```
kalinkaaaa14, line 2:
```suggestion
	ROOT: "/",
```

apps/frontend/src/libs/components/modal/modal.tsx
```diff
@@ -0,0 +1,67 @@
...
+	isOpen,
+	onClose,
+}: Properties): React.JSX.Element => {
+	useEffect((): (() => void) => {
+		if (!isOpen) {
+			return () => {};
+		}
+
+		const onKeyDown = (event: KeyboardEvent): void => {
+			if (event.key === "Escape") {
```
kalinkaaaa14, line 25:
```suggestion
			if (event.key === KeyboardKey.ESCAPE) {
```

kalinkaaaa14 (changes requested):

apps/frontend/src/pages/dashboard/dashboard.tsx
vldkorin: > 

I did that strange things, because we need to send 
```ts
coordinates: payload.location.coordinates
```
as strings
And then they are converted into numbers

longitudeSchema = z.string().transform(parseToFloat).pipe(z.number().min(...).max(...))

latitudeSchema = z.string().transform(parseToFloat).pipe(z.number().min(...).max(...))

locationSchema = z.object({ type: z.literal("Point"), coordinates: z.tuple([longitudeSchema, latitudeSchema]) })

And also i think we should fix swagger
image
image

liza-veis: Is the problem with converting a number to a string? let's keep coordinates to string on frontend and they should be strings in swagger too

packages/shared/src/modules/points-of-interest/libs/enums/points-of-interest-validation-rule.ts
```diff
@@ -1,4 +1,5 @@
 const PointsOfInterestValidationRule = {
+	DESCRIPTION_MAX_LENGTH: 100,
```
liza-veis, line 2: let's make it larger and also add it to the migration string(1000)
```suggestion
	DESCRIPTION_MAX_LENGTH: 1000,
```

kalinkaaaa14:
@vldkorin as we have already merged the map, what do you think about adding it in this Modal with location search and implementing the full flow of POI creation? If you think it will be too many changes, we can create a separate ticket for that

vldkorin:
> @vldkorin as we have already merged the map, what do you think about adding it in this Modal with location search and implementing the full flow of POI creation? If you think it will be too many changes, we can create a separate ticket for that

For me, it's better to have separate ticket

liza-veis:
> For me, it's better to have separate ticket
Good, I will create a separate ticket then

apps/frontend/src/pages/dashboard/dashboard.tsx
liza-veis: Yes, let's update it to use strings everywhere, map-provider and map-client are our custom modules, so it's ok to do whatever conversion we need there

vldkorin: > Yes, let's update it to use strings everywhere, map-provider and map-client are our custom modules, so it's ok to do whatever conversion we need there

So, i haven't noticed that but i need to update every place where it's used(poi module, route and other )

apps/frontend/src/libs/components/create-poi-modal/create-poi-modal.tsx
```diff
@@ -0,0 +1,104 @@
...
+						errors={errors}
+						label="Description (optional)"
+						name="description"
+					/>
+				</div>
+
+				<div className={styles["footer"]}>
+					<div className="button">
+						<Button label="Create" type="submit" />
+						<div />
```
lavandosya, line 96: do we need this div?
```suggestion
```

apps/frontend/src/libs/components/create-poi-modal/libs/types/form-field.type.ts
lavandosya, line 1: The file name and exported type should match

apps/frontend/src/libs/components/create-poi-modal/libs/types/local-poi-request-dto.type.ts
lavandosya, line 1: The file name and exported type should match

apps/frontend/src/libs/components/create-poi-modal/styles.module.css
```diff
@@ -0,0 +1,34 @@
+.header {
+	display: flex;
+	align-items: center;
+	justify-content: space-between;
+	margin: 0 0 12px;
```
lavandosya, line 5:
```suggestion
	margin-bottom: 0 0 12px;
```

apps/frontend/src/libs/components/create-poi-modal/styles.module.css
```diff
@@ -0,0 +1,34 @@
...
+
+.field {
+	display: flex;
+	flex-direction: column;
+	gap: 8px;
+}
+
+.footer {
+	display: flex;
+	gap: 8px;
```
lavandosya, line 28:
```suggestion
```

apps/frontend/src/libs/components/create-poi-modal/create-poi-modal.tsx
```diff
@@ -0,0 +1,104 @@
...
+					<TextArea
+						control={control}
+						errors={errors}
+						label="Description (optional)"
+						name="description"
+					/>
+				</div>
+
+				<div className={styles["footer"]}>
+					<div className="button">
```
lavandosya, line 94:
```suggestion
					<div className={styles["button"]}> or remove
```

lavandosya (changes requested):

apps/backend/src/db/migrations/20250812102347_add_description_to_points_of_interest.ts
```diff
@@ -0,0 +1,18 @@
+import { type Knex } from "knex";
+
+const TABLE_NAME = "points_of_interest";
+const DESCRIPTION_LENGTH = 1000;
```
liza-veis, line 4: I thought for a bit and understood that 1000 chars is actually very very low 😅 , let's use 8000

apps/frontend/src/pages/dashboard/dashboard.tsx
liza-veis: Oh, I seem to understand what the original issue was. The validation schemas for long and lat for coordinates were done originally for strings because it's used in search queries where we use only strings, and here we don't need to convert anything because we pass it in the body, so we can just pass numbers. I got confused a bit and thought that maybe we store those coordinates as strings on the backend, which is also ok. 

In our case, the issue seems to be in the shared validation schemas for lat lon that don't correspond usual format (number). I never actually liked re-using and sharing validation schemas, but thought maybe it would be ok for lat / lon, and now we have a conflict in data formats 😅 The correct way to handle this would be to define a schema for lon / lat separately our common lon / lat in a usual format - just numbers, without conversion. And since we see that we definitely need separate validation, let's remove those lat /lon schemas as reusable ones, integrate the existing ones into the query schema, and define the validation for our current schema separately. We definitely don't want to convert datatypes in an unnecessary way, just to satisfy the shared validation schema.

liza-veis: CC @vyacheslavKhoroshev Just FYI, I think it can be useful for your experience to know that we had this case and why we usually prefer to avoid shared validation schemas. I haven't anticipated this issue either 😁

vyacheslavKhoroshev:
i checked POI search, it doesn't work with lat&lng as strings. So you should change type to number before it goes to the repository. i mean findNearby method. https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/issues/188 if you going to change validate schema.

the main idea, not to break other logic)))

apps/frontend/src/pages/dashboard/dashboard.tsx
```diff
@@ -31,12 +47,29 @@ const Dashboard = (): React.JSX.Element => {
...
 	});
 
+	const [isCreateOpen, setIsCreateOpen] = useState<boolean>(false);
+
+	const handleOpen = useCallback(() => {
+		setIsCreateOpen(true);
+	}, []);
+	const handleClose = useCallback(() => {
+		setIsCreateOpen(false);
+	}, []);
```
kalinkaaaa14, line 60: The state could me managed in 1 handler

apps/frontend/src/pages/dashboard/dashboard.tsx
```diff
@@ -31,12 +47,29 @@ const Dashboard = (): React.JSX.Element => {
...
 
 	const { control } = useAppForm<FormValues>({
-		defaultValues: {
-			multiColors: [],
-			singleColor: null,
-		},
+		defaultValues: { multiColors: [], singleColor: null },
 	});
 
+	const [isCreateOpen, setIsCreateOpen] = useState<boolean>(false);
```
kalinkaaaa14, line 53:
```suggestion
	const [isCreatePOIOpen, setCreatePOIOpen] = useState<boolean>(false);
```

apps/frontend/src/libs/components/modal/modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+		};
+	}, [isOpen, onClose]);
+
+	const handleBackdropClick = useCallback((): void => {
+		onClose();
+	}, [onClose]);
+
+	if (!isOpen) {
+		return <></>;
+	}
```
kalinkaaaa14, line 44:
```suggestion
	if (!isOpen) {
		return null;
	}
```

vldkorin: Type 'null' is not assignable to type 'Element'

apps/frontend/src/pages/dashboard/dashboard.tsx
vldkorin: > Держава могла б мені впоратися в 1 хендлера

yes, but 
```
JSX props should not use arrow functions eslint
```
that's why it was done

liza-veis (approved):
LGTM 😎 , please fix this comment https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/299/files#r2271294592 and this one https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/299#discussion_r2274769240

apps/frontend/src/pages/dashboard/dashboard.tsx
liza-veis: I think what Alina meant that instead of having handleOpen/handleClose, we can have handleModalToggle or something like this nand just change the value to the opposite one

vldkorin: > handleModalToggle

like that?
```ts
	const handleModalToggle = useCallback(() => {
		setIsCreatePOIOpen((previous) => !previous);
	}, []);
```

vldkorin:
> LGTM 😎 , please fix this comment https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/299/files#r2271294592 and this one [#299 (comment)](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/299#discussion_r2274769240)

About first comment, without that wrapper button stretches across the entire container

lavandosya:
> > LGTM 😎 , please fix this comment https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/299/files#r2271294592 and this one [#299 (comment)](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/299#discussion_r2274769240)
> 
> About first comment, without that wrapper button stretches across the entire container so it becomes too wide and out of place

You just can remove button classname and left a div
Screenshot 2025-08-14 at 11 21 04

lavandosya:
Except from this looks good to me 💫

lavandosya (approved):

apps/frontend/src/pages/dashboard/dashboard.tsx
```diff
@@ -31,12 +47,26 @@ const Dashboard = (): React.JSX.Element => {
...
 	});
 
+	const [isCreatePOIOpen, setIsCreatePOIOpen] = useState<boolean>(false);
+
+	const handleModalToggle = useCallback(() => {
+		setIsCreatePOIOpen((previous) => !previous);
+	}, []);
+	const handleSubmit = useCallback(
+		async (payload: PointsOfInterestRequestDto) => {
+			const result = await dispatch(poiActions.create(payload));
```
liza-veis, line 60: Check these two comments:
1. We can't await the dispatch result - https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/252#discussion_r2259874779
2. We need to trigger toast notification inside action creator: https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/363#discussion_r2274778328

To setIsCreatePOIOpen to false, you need to create a property in store that will track creation status and add useEffect here which will close the module on state being fulfilled

packages/shared/src/modules/points-of-interest/libs/validation-schemas/location.validation-schema.ts
```diff
@@ -1,11 +1,30 @@
...
+	.min(
+		PointsOfInterestValidationRule.LONGITUDE_MIN,
+		PointsOfInterestValidationMessage.LONGITUDE_MIN,
+	)
+	.max(
+		PointsOfInterestValidationRule.LONGITUDE_MAX,
+		PointsOfInterestValidationMessage.LONGITUDE_MAX,
+	);
+
+const latitudeSchema = z
```
liza-veis, line 18: let's get rid of common schemas and define them directly

liza-veis (changes requested):

liza-veis (approved):

Issue #268, opened 2025-08-08
feat: Implement Route description page


liza-veis:
### What feature?

- [ ] Route description page design (For now, only name, description, and use static pre-defined images)
- [ ] Render appropriate page on `app/route/:id` route
- [ ] `app/route/:id` should be a public route (accessible without token)
- [ ] In case of rejection on the backend side, redirect to the NotFound page

Note: If any adjustments on the backend in needed, this should be covered in this task.

We also have a similar ticket for POI description page #245, make sure that the design and implementation are the same, collaborate with @h4vefundieyoung if needed

### Screenshots

<img width="1280" height="1300" alt="Image" src="https://github.com/user-attachments/assets/1513f8d1-8d48-4583-b9ed-21ca4f134100" />

Anna-Kutsenko:
<img width="1575" height="841" alt="Image" src="https://github.com/user-attachments/assets/fb3e2e2e-6ba3-4c3d-9256-41bace97a1bd" />

PR #337, opened 2025-08-12
feat(frontend/backend/shared): implement points of interest details page ss-245

h4vefundieyoung:
Updates:

- Added description field to the points of interest table.
- All the poi-table related logic has been updated with the description prop.
- Points of interest module with required reducers/actions has been implemented on the client side.
- `/points-of-interest/:id` route has been added on the client side.

Currently waiting for #268 to be resolved so that route could be added to the white list routes.
Screenshot 2025-08-12 at 14 43 59


apps/backend/src/db/migrations/20250812105948_add_poi_description_column.ts
```diff
@@ -0,0 +1,17 @@
...
+
+function down(knex: Knex): Promise<void> {
+	return knex.schema.alterTable(TABLE_NAME, (table) => {
+		table.dropColumn("description");
+	});
+}
+
+function up(knex: Knex): Promise<void> {
+	return knex.schema.alterTable(TABLE_NAME, (table) => {
+		table.string("description");
```
vldkorin, line 13:
```suggestion
table.string("description",1000).nullable();

Comment (vldkorin) [reply apps/backend/src/db/migrations/20250812105948_add_poi_description_column.ts:13]: > 

I think we shouldn't do the same thing, and that's also about  request/response dto and validation schemes. You can check out my pr #299

File: apps/frontend/src/assets/images/mocks/poi2.jpg
Comment (timosh-yura) [inline apps/frontend/src/assets/images/mocks/poi2.jpg:1]: please export images without border-radius or copy from
https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/283/files

File: apps/frontend/src/modules/points-of-interest/slices/points-of-interest-details.slice.ts [hunk 2]
```diff
@@ -0,0 +1,45 @@
...
+		builder.addCase(loadById.fulfilled, (state, action) => {
+			const incomingPoiDetails = action.payload.data;
+			state.dataStatus = DataStatus.FULFILLED;
+			state.pointsOfInterestDetails = [
+				...state.pointsOfInterestDetails.filter(
+					(poiDetail) => poiDetail.id !== incomingPoiDetails.id,
+				),
+				incomingPoiDetails,
+			];
+		});
```
timosh-yura, line 32: I`m not sure if it should accumulate pois instead of just replacing, if this store is used for displaying info about an individual poi

apps/frontend/src/pages/points-of-interest-details/points-of-interest-details.tsx
```diff
@@ -0,0 +1,75 @@
...
+		({ id }) => id === formattedId,
+	);
+
+	useEffect(() => {
+		if (formattedId) {
+			void dispatch(actions.loadById(formattedId));
+		} else {
+			void (async (): Promise<void> => {
+				try {
+					await navigate(AppRoute.SIGN_IN, { replace: true });
```
timosh-yura, line 37: it should redirect to NotFound page

apps/backend/src/db/migrations/20250812105948_add_poi_description_column.ts
h4vefundieyoung: @timosh-yura do we have it defined somewhere, i haven't seen such things

apps/frontend/src/pages/points-of-interest-details/points-of-interest-details.tsx
h4vefundieyoung: it definetely should, lol

apps/backend/src/db/migrations/20250812105948_add_poi_description_column.ts
h4vefundieyoung: let's ask @liza-veis

apps/backend/src/db/migrations/20250812105948_add_poi_description_column.ts
```diff
@@ -0,0 +1,24 @@
...
+async function up(knex: Knex): Promise<void> {
+	await knex.transaction(async (trx) => {
+		await trx.schema.alterTable(TABLE_NAME, (table) => {
+			table.text("description");
+		});
+		await trx
+			.table(TABLE_NAME)
+			.whereNull("description")
+			.update({ description: DEFAULT_DESCRIPTION });
+	});
```
kalinkaaaa14, line 21: I have doubts that this migration is working actually, or am I missing something? By default columns are non-nullable. So, if your table has already data inserted - and it should already have - then this migration should fail on the attempt to add a required field, as it doesn't allow null values. 

In the real world projects it's a usual situation that you need to extend the existing table with some new columns, and, for example if this column is required - then the flow should be like this - add new column, mark it nullable - seed default data - make this column non nullable after that. Hope that was helpful to understand

In our case description is an optional field, so we don't need at all this manipulation, so please remove DEFAULT_DESCRIPTION and make description column nullable.

apps/backend/src/db/migrations/20250812190635_add_poi_description_required_modifier.ts
```diff
@@ -0,0 +1,17 @@
...
+	return knex.schema.alterTable(TABLE_NAME, (table) => {
+		table.text("description").nullable().alter();
+	});
+}
+
+function up(knex: Knex): Promise<void> {
+	return knex.schema.alterTable(TABLE_NAME, (table) => {
+		table.text("description").notNullable().alter();
+	});
+}
```
kalinkaaaa14, line 15: You don't need to have two separate migrations. Changes are very small and they are related, better to do in 1 migration

apps/backend/src/modules/points-of-interest/points-of-interest.controller.ts
```diff
@@ -43,12 +43,16 @@ import {
...
  *         - name
+ *         - description
  *       properties:
  *         location:
  *           $ref: '#/components/schemas/PointsOfInterestLocation'
  *         name:
  *           type: string
  *           example: "Central Park"
+ *         description:
+ *           type: string
```
kalinkaaaa14, line 54:
```suggestion
 *           type: string | null
```

apps/backend/src/modules/points-of-interest/points-of-interest.controller.ts
```diff
@@ -65,6 +69,9 @@ import {
...
```
kalinkaaaa14, line 73:
```suggestion
 *           type: string | null
```

apps/backend/src/modules/points-of-interest/points-of-interest.entity.ts
```diff
@@ -1,68 +1,81 @@
 import { type Entity, type PointGeometry } from "~/libs/types/types.js";
 
 class PointsOfInterestEntity implements Entity {
+	private description: string;
```
kalinkaaaa14, line 4:
```suggestion
	private description: string | null;
```

apps/backend/src/modules/points-of-interest/points-of-interest.model.ts
```diff
@@ -9,6 +9,7 @@ class PointsOfInterestModel extends AbstractModel {
 		return DatabaseTableName.POINTS_OF_INTEREST;
 	}
 
+	public description!: string;
```
kalinkaaaa14, line 12:
```suggestion
	public description!: string | null;
```

apps/frontend/src/index.tsx
```diff
@@ -28,6 +29,10 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
 				routes={[
 					{
 						children: [
+							{
+								element: <PointsOfInterestDetails />,
+								path: `${AppRoute.POINTS_OF_INTEREST}/:id`,
```
kalinkaaaa14, line 34: Add '/points-of-interest/:id' to AppRoute

apps/frontend/src/modules/points-of-interest/slices/actions.ts
```diff
@@ -0,0 +1,18 @@
+import { createAsyncThunk } from "@reduxjs/toolkit";
+
+import { type APIResponse, type AsyncThunkConfig } from "~/libs/types/types.js";
+import { type PointsOfInterestResponseDto } from "~/modules/points-of-interest/points-of-interest.js";
+
+import { name as sliceName } from "./points-of-interest-details.slice.js";
+
+const loadById = createAsyncThunk<
```
kalinkaaaa14, line 8: better to call getById I think, more intuitive

apps/frontend/src/modules/points-of-interest/slices/points-of-interest-details.slice.ts
kalinkaaaa14: I think Yurii's point was that this slice should be used only for details page, and here we will have only 1 pointOfInterestDetails, not the array. And I agree with this point, for places where we will be storing array of points we will have another slices

apps/backend/src/db/migrations/20250812190635_add_poi_description_required_modifier.ts
h4vefundieyoung: it's not about size. as i know it's just a bad practice to describe multiple alters in 1 migration.
and well, it's just useless if the field should remains optional.

apps/frontend/src/index.tsx
h4vefundieyoung: @kalinkaaaa14 i wanted at first, but well, won't we have this general points of interest route?

apps/frontend/src/modules/points-of-interest/slices/actions.ts
h4vefundieyoung: @kalinkaaaa14 agree. but well, i picked that naming pattern from the user module, i believe @liza-veis defined it. if it's ok - i'll rename it.

apps/backend/src/db/migrations/20250812105948_add_poi_description_column.ts
liza-veis: @vldkorin, thanks for mentioning. I agree that we want to apply a string limit here. When I checked what it would look like with 1000 characters, I understood it would be too small, and I think we need to update all descriptions to something like 8000 characters. The user may possibly have large descriptions with markups for routes/POIs, so let's update this

apps/frontend/src/assets/images/mocks/poi2.jpg
liza-veis: It still makes sense, even mock data should follow application format and conventions, because we never know when we will actually get rid of them

apps/backend/src/db/migrations/20250812105948_add_poi_description_column.ts
kalinkaaaa14: Oh yeah I see, seems I a bit mixed up. Then yeah, It should work with existing data

apps/backend/src/db/migrations/20250812190635_add_poi_description_required_modifier.ts
kalinkaaaa14: I don't know where did you find this information you rely on, but this kind of stuff in the real project usually is done in 1 the migration, and all is covered with a transaction, so if something goes wrong - previous changes would be rollbacked

apps/frontend/src/index.tsx
kalinkaaaa14: maybe we will, but we should have 2 separate values in AppRoute enum for that

apps/frontend/src/modules/points-of-interest/points-of-interest.api.ts
```diff
@@ -0,0 +1,36 @@
...
+import { PointsOfInterestApiPath } from "./libs/enums/enums.js";
+import { type PointsOfInterestResponseDto } from "./libs/types/types.js";
+
+type Constructor = {
+	baseUrl: string;
+	http: HTTP;
+	storage: Storage;
+};
+
+class PointsOfInterestApi extends BaseHTTPApi {
```
mxm-kotyk, line 16: Shouldn't it be `PointOfInterestApi` (singular)?

mxm-kotyk: Yeah, you are right. It's just strange that we have `UserApi` in singular. Anyway, it's not a big deal 😅.

vldkorin: When resolving conflicts, use the plural form everywhere

h4vefundieyoung: @vldkorin [https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/283#discussion_r2276618192](convetion)

apps/frontend/src/assets/images/mocks/poi1.jpg
liza-veis, line 1: Let's use the images that @timosh-yura used for routes page, instead of adding a new one

apps/frontend/src/pages/points-of-interest-details/points-of-interest-details.tsx
```diff
@@ -0,0 +1,82 @@
...
+				try {
+					await navigate(AppRoute.NOT_FOUND, { replace: true });
+				} catch {
+					toastNotifier.showError(
+						CommonExceptionMessage.COMMON_EXCEPTION_MESSAGE,
+					);
+				}
+			})();
+		}
+	}, [dispatch, formattedId, navigate]);
```
liza-veis, line 42: Take a look at how it's implemented in the routes details page, and here we need fully the same logic.

As for the current approach itself:
All side-effect logic should happen in action creators or middlewares outside of components; we can't await side effects or call other services. 

The architecture is as follows:

In ui components, we dispatch an action using action creators 
-> all side effects logic happens in the action creator 
-> store updates the state 
-> ui components listen to state updates and react to them.

It means we can't use toastNotification or await navigate, and we can't await dispatch either.

apps/frontend/src/pages/points-of-interest-details/styles.module.css
```diff
@@ -0,0 +1,67 @@
...
+		flex-direction: row;
+		justify-content: center;
+		width: 100%;
+	}
+}
+
+@media (width <= 640px) {
+	.poi-header {
+		line-height: 1;
+	}
```
liza-veis, line 66: I think there is a point in changing font size on smaller screens, but not font height. If what you are trying to achieve is having less space between the lines when the text wraps, this should be the initial line height. I see you used line-height:3 as the initial value, and I don't understand what we need this for, since it definitely should be somewhere between 1 and 1.4 for headers, so that they look ok when they wrap, even on large screens
```suggestion
```

apps/frontend/src/pages/points-of-interest-details/styles.module.css
liza-veis, line 1: Make sure the styles here are the same as for the route-details page, we don't want discrepancies

apps/frontend/src/pages/points-of-interest-details/points-of-interest-details.tsx
```diff
@@ -0,0 +1,82 @@
...
+	return (
+		<>
+			{isRejected && <Navigate replace to={AppRoute.NOT_FOUND} />}
+			<main className={styles["poi-container"]}>
+				{pointsOfInterestDetails && (
+					<>
+						<h2 className={styles["poi-header"]}>
+							{pointsOfInterestDetails.name}
+						</h2>
+						<div className={styles["poi-gallery"]}>
```
liza-veis, line 53: We already have an implementation for the image gallery for the route details page. Let's move it to the common components and use it here

liza-veis (changes requested):

apps/frontend/src/modules/points-of-interest/points-of-interest-api.ts
```diff
@@ -36,6 +36,20 @@ class PointOfInterestApi extends BaseHTTPApi {
...
 		return (await response.json()) as APIResponse<PointsOfInterestResponseDto>;
 	}
+
+	public async getById(
+		id: number,
+	): Promise<APIResponse<PointsOfInterestResponseDto>> {
+		const response = await this.load<APIResponse<PointsOfInterestResponseDto>>(
+			this.getFullEndpoint(PointsOfInterestApiPath.ROOT, String(id), {}),
+			{
+				hasAuth: true,
```
kalinkaaaa14, line 46: It should work without authorization, isn't it?
@liza-veis

apps/backend/src/modules/points-of-interest/points-of-interest.service.ts
```diff
@@ -98,7 +98,7 @@ class PointsOfInterestService implements Service {
 		id: number,
 		payload: PointsOfInterestRequestDto,
 	): Promise<PointsOfInterestResponseDto> {
-		const { description, location, name } = payload;
+		const { description = null, location, name } = payload;
```
kalinkaaaa14, line 101: I think there is no point of adding the default value for the description here 🤔  It would make sense if description was optional field ( meaning described in the type like description?: string | null ), but in our case undefined is not possible

apps/backend/src/modules/points-of-interest/points-of-interest.controller.ts
kalinkaaaa14: Should be fixed, seems to be missed

apps/backend/src/modules/points-of-interest/points-of-interest.controller.ts
kalinkaaaa14: Should be fixed, seems to be missed

apps/frontend/src/modules/points-of-interest/points-of-interest-api.ts
h4vefundieyoung: @kalinkaaaa14 it should

liza-veis (changes requested):

PR #343, opened 2025-08-12
fix(frontend): add slide transition animation ss-281

DenisDemirov:
Added slide transition animation
Recording 2025-08-12 162734


apps/frontend/src/libs/components/sidebar/styles.module.css
```diff
@@ -13,15 +13,28 @@
 
 .drawer {
 	z-index: 1;
-	width: max-content;
+	width: 74px;
```
timosh-yura, line 16: I'm not sure if it's necessary
Except that, LGTM

kalinkaaaa14:
On the video it looks like the page is blinking during sidebar animation, is it a problem with a video, or this problem does exist?

DenisDemirov:
> On the video it looks like the page is blinking during sidebar animation, is it a problem with a video, or this problem does exist?

Yes, its some artifact after covert to gif

apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -74,7 +74,7 @@ const Sidebar = ({ navigationItemsGroups }: Properties): React.JSX.Element => {
 								<div
 									className={combineClassNames(
 										styles["title"],
-										!isOpen && "visually-hidden",
+										!isOpen && "hidden",
```
liza-veis, line 77: Is this class repeas our  "visually-hidden" class? If so, let's use visually-hidden, it was specifically created as a utility class for such cases

DenisDemirov: Not fully, i in "hidden" i removed height : 1px to make animation for hidden elements more smooth, without it container change height during panel opening

kalinkaaaa14 (approved):

liza-veis (approved):

PR #346, opened 2025-08-12
feat(frontend): image upload ss-220

### LINKED ISSUE #220: feat: Add file uploading functionality
### What feature?

Implement file uploading functionality using [AWS S3](https://aws.amazon.com/s3/)

- [x] Create files table
```
Table files {
  id integer [primary key]
  created_at timestamp
  updated_at timestamp
  url varchar
  content_type varchar (e.g. 'image/jpeg')
}
```
- [x] Add a separate module for file uploading to AWS
- [x] Add files module with the needed logic
- [x] Add `POST /files` route for file creation
- [ ] Image retrieval should be implemented using presigned URLs
- [x] Add files fastify plugin
  - [x] Limit file extension to jpeg/png
  - [x] file size: < 10Mb

### Screenshots

_No response_

vyacheslavKhoroshev:
- file API;
- file module;
- aws module;
- miltipart plugin:
- fix carausel component;
- update dashdoard page;

Resolves #220


vyacheslavKhoroshev:
> I believe fixing carousel component doesn`t relate to the tiket.
> Generally LGTM

these were small and even necessary fixes. because I realized that there was a problem with large files.

apps/backend/src/libs/modules/aws/base-aws.module.ts
```diff
@@ -0,0 +1,58 @@
+import { PutObjectCommand, S3Client } from "@aws-sdk/client-s3";
+
+import { type Config } from "~/libs/modules/config/config.js";
+import { HTTPCode } from "~/libs/modules/http/http.js";
+import { type Logger } from "~/libs/modules/logger/logger.js";
+import { type FileContentType } from "~/modules/files/files.js";
+
+import { AWSError } from "./libs/exeptions/exeptions.js";
+
+class AWSService {
```
kalinkaaaa14, line 10: let's call it AWSFileService, as AWS have a lot of services and from the naming it's not clear which one. And do not forget to change filename to file.service.ts

apps/backend/src/libs/modules/aws/base-aws.module.ts
```diff
@@ -0,0 +1,58 @@
...
+			});
+
+			await this.s3Client.send(command);
+
+			const url = `https://${this.bucketName}.s3.amazonaws.com/${key}`;
+			this.logger.info(`File uploaded to AWS S3: ${key}`);
+
+			return url;
+		} catch (error) {
+			const message = error instanceof Error ? error.message : String(error);
```
kalinkaaaa14, line 48: let's check whether error is instace of S3ServiceException

apps/backend/src/libs/modules/aws/libs/exeptions/aws.exeption.ts
```diff
@@ -0,0 +1,5 @@
+import { HTTPError } from "@smartscapes/shared";
+
+class AWSError extends HTTPError {}
```
kalinkaaaa14, line 3: let's name it FilesError

apps/backend/src/modules/files/files.service.ts
```diff
@@ -0,0 +1,79 @@
...
+
+import { FilesEntity } from "./files.entity.js";
+import { type FilesRepository } from "./files.repository.js";
+import {
+	type FileContentType,
+	type FileUploadRequestDto,
+	type FileUploadResponseDto,
+} from "./libs/types/types.js";
+
+class FilesService implements Service {
```
kalinkaaaa14, line 14:
```suggestion
class FileService implements Service {
```

apps/backend/src/modules/files/files.service.ts
```diff
@@ -0,0 +1,79 @@
...
+	private awsService: AWSService;
+	private filesRepository: FilesRepository;
```
kalinkaaaa14, line 16:
```suggestion
	private fileRepository: FileRepository;
```

packages/shared/src/modules/files/libs/types/file-upload-request-dto.type.ts
```diff
@@ -0,0 +1,9 @@
+import { type ValueOf } from "../../../../libs/types/types.js";
+import { type FileFolderName } from "../enums/enums.js";
+
+type FileUploadRequestDto<T> = {
+	file: T;
```
kalinkaaaa14, line 5: let's type it like this. It's ok for small & medium size images, and that should be our case

file: Buffer;

packages/shared/src/modules/files/libs/enums/files-validation-message.enum.ts
```diff
@@ -0,0 +1,9 @@
+import { FileFolderName } from "./file-folder-name.enum.js";
+
+const FilesValidationMessage = {
+	INVALID_FOLDER_NAME: `Invalid folder name. Allowed values: ${Object.values(
+		FileFolderName,
+	).join(", ")}`,
+} as const;
```
kalinkaaaa14, line 7: let's create a separate constant FileFolderNameValues = Object.values(FileFolderName); and reuse it here in the enum

packages/shared/src/modules/files/libs/enums/file-content.enum.ts
```diff
@@ -0,0 +1,7 @@
+const FileContent = {
```
kalinkaaaa14, line 1: FileMimeType

packages/shared/src/libs/constants/constants.ts
```diff
@@ -1,2 +1,3 @@
+export { BYTES_IN_MB } from "./bytes-in-mb.constants.js";
```
kalinkaaaa14, line 1:
```suggestion
export { BYTES_IN_MB } from "./bytes-in-mb.constant.js";
```

apps/frontend/src/pages/dashboard/dashboard.tsx
```diff
@@ -48,11 +97,26 @@ const Dashboard = (): React.JSX.Element => {
...
+						/>
+						<input
+							accept="image/*"
+							disabled={isUploading}
+							onChange={handleFileUpload}
+							ref={fileInputReference}
+							style={{ display: "none" }}
+							type="file"
+						/>
+					</label>
```
kalinkaaaa14, line 119: Let's add file upload to Profile page?

apps/backend/src/libs/modules/plugins/multipart/multipart.plugin.ts
```diff
@@ -0,0 +1,65 @@
...
+				throw new FilesError({
+					message: FilesExceptionMessage.FILE_REQUIRED,
+					status: HTTPCode.BAD_REQUEST,
+				});
+			}
+
+			const ALLOWED_FILE_TYPES = Object.values(
+				FileContent,
+			) as readonly string[];
+			const IS_INVALID_TYPE = !ALLOWED_FILE_TYPES.includes(file.mimetype);
```
kalinkaaaa14, line 47:
```suggestion
			const isInvalidType = !ALLOWED_FILE_TYPES.includes(file.mimetype);
```

apps/frontend/src/pages/dashboard/dashboard.tsx
vyacheslavKhoroshev: I have already attached it to task.  [video](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/issues/220#issuecomment-3180366740)

packages/shared/src/modules/files/libs/types/file-upload-request-dto.type.ts
kalinkaaaa14: Hmm, I need to check it, but what I thought is that when you are uploading file on the frontend - it should be of type File, which is basically a Blob already, and you can do something like  

formData.append('file', selectedFile);

and then while registering plugin

app.register(fastifyMultipart, {
  limits: {
    fileSize: value
  },
  attachFieldsToBody: true,
  throwFileSizeLimit: false,
});

kalinkaaaa14: oh seems I understood, when uploading the file we can add a type where basically file will be a formdata type

file: FormData;
  
and then on the plugin on backend convert it to Buffer type and add to the request fileBuffer field, and then those conversion to buffer on service layer that you have now would be redundant

vyacheslavKhoroshev: yes, i tried like this, i mean 'attachFieldsToBody: true' or smth like this. but had problems. i will try one more time, because i dont remember where the problem is.

apps/backend/src/libs/modules/aws/base-aws-file.module.ts
```diff
@@ -0,0 +1,64 @@
...
+			return url;
+		} catch (error) {
+			const message =
+				error instanceof S3ServiceException ? error.message : String(error);
+			this.logger.error(`Failed to upload file to AWS S3: ${message}`);
+
+			throw new AWSFileUploadError({
+				message,
+				status: HTTPCode.INTERNAL_SERVER_ERROR,
+			});
```
kalinkaaaa14, line 59: The moment to improve here also - status code 500. It's not the best idea to throw internal server errors, if it's handled exception. Can we receive status code from the AWS error? And if the error is not instance of S3ServiceException - let's return 404 and some default message, like 'The problem occured while uploading the file', because currently we will show some not readable message to user making String(error), where error could be probably an object. And the exact error what has actually happened should be present in the logs, not in user message

apps/backend/src/libs/modules/aws/libs/exeptions/exeptions.ts
```diff
@@ -0,0 +1 @@
+export { AWSFileUploadError } from "./aws-file.exeption.js";
```
kalinkaaaa14, line 1:
```suggestion
export { AWSFileUploadError } from "./aws-file-upload.exception.js";
```

apps/frontend/src/pages/dashboard/dashboard.tsx
liza-veis: I'm ok with it, but right after the demo, we will need to remove this. Also, all side effects should happen through redux actions; we can't just call the service in other components. In this case, let's leave it as it is already to show for the demo. But our code will soon be checked against quality criteria.

vldkorin:
Except for the changes you need to make, looks good

apps/backend/src/libs/modules/aws/base-aws-file.module.ts
vyacheslavKhoroshev: > The problem occured while uploading the file

image

error when wrong credentials

packages/shared/src/modules/files/libs/types/file-upload-request-dto.type.ts
vyacheslavKhoroshev: @kalinkaaaa14  did it! 
i wanted to catch errors in plugin but could not resolve how.
that is why checked validation mime in servise and caught error of big size file in error handler.

liza-veis (changes requested):

apps/backend/src/libs/modules/server-application/libs/helpers/is-fastify-error.helper.ts
```diff
@@ -0,0 +1,12 @@
+const isFastifyError = (error: unknown): boolean => {
```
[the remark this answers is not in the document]
vyacheslavKhoroshev: wow! I was wondering how to catch that elegantly, I will do it in the morning, thanks.

packages/shared/src/modules/files/libs/enums/enums.ts
```diff
@@ -0,0 +1,3 @@
+export { FileFolderName } from "./file-folder-name.enum.js";
+export { FilesApiPath } from "./files-api-path.enum.js";
+export { FilesValidationMessage } from "./files-validation-message.enum.js";
```
kalinkaaaa14, line 3:
```suggestion
export { FileValidationMessage } from "./file-validation-message.enum.js";
```

apps/frontend/src/modules/files/files-api.ts
```diff
@@ -0,0 +1,57 @@
...
+	type FileUploadResponseDto,
+} from "./libs/types/types.js";
+
+type Constructor = {
+	baseUrl: string;
+	http: HTTP;
+	storage: Storage;
+};
+
+class FilesApi extends BaseHTTPApi {
```
kalinkaaaa14, line 19:
```suggestion
class FileApi extends BaseHTTPApi {
```

apps/backend/src/modules/files/libs/enums/files-exeption-message.enum.ts
```diff
@@ -0,0 +1,8 @@
+const FilesExceptionMessage = {
```
kalinkaaaa14, line 1:
```suggestion
const FileExceptionMessage = {
```

apps/backend/src/modules/files/files.repository.ts
```diff
@@ -0,0 +1,37 @@
+import { type Repository } from "~/libs/types/types.js";
+
+import { FilesEntity } from "./files.entity.js";
+import { type FilesModel } from "./files.model.js";
+
+const FILES_LIMIT = 10;
+
+class FileRepository implements Repository {
+	private filesModel: typeof FilesModel;
```
kalinkaaaa14, line 9:
```suggestion
	private fileModel: typeof FileModel;
```

apps/backend/src/modules/files/files.model.ts
```diff
@@ -0,0 +1,17 @@
+import {
+	AbstractModel,
+	DatabaseTableName,
+} from "~/libs/modules/database/database.js";
+
+import { type FileMimeType } from "./libs/types/types.js";
+
+class FilesModel extends AbstractModel {
```
kalinkaaaa14, line 8:
```suggestion
class FileModel extends AbstractModel {
```

apps/backend/src/modules/files/files.entity.ts
```diff
@@ -0,0 +1,75 @@
+import { type Entity } from "~/libs/types/types.js";
+
+import { type FileMimeType } from "./libs/types/types.js";
+
+class FilesEntity implements Entity {
```
kalinkaaaa14, line 5:
```suggestion
class FileEntity implements Entity {
```

apps/backend/src/libs/modules/server-application/libs/helpers/is-fastify-error.helper.ts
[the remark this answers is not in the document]
vyacheslavKhoroshev: Do we have this class? Or I should create own? I will figure out...

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -170,6 +172,23 @@ class BaseServerApplication implements ServerApplication {
 					return reply.status(error.status).send(response);
 				}
 
+				if (isFastifyError(error)) {
```
kalinkaaaa14, line 175: If you decide to leave this helper, than it should be named checkIsFastifyError
```suggestion
				if (checkIsFastifyError(error)) {
```

apps/backend/src/libs/modules/aws/libs/exeptions/aws-file-upload.exception..ts
kalinkaaaa14, line 1: apps/backend/src/libs/modules/aws/libs/exeptions/aws-file-upload.exception.ts

apps/backend/src/libs/modules/server-application/libs/helpers/is-fastify-error.helper.ts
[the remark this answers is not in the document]
kalinkaaaa14: I think Liza meant here that this type exists from fastify/multipart, no need to create your own

packages/shared/src/modules/files/libs/types/file-upload-request-dto.type.ts
kalinkaaaa14: It could be handled in the plugin, that's a usual place where such things are handled actually, and I think it can be done if you change  throwFileSizeLimit: false, and you need to add manual check if the file was truncated. Like this

const { file } = request.body;
 
if (file.truncated) {
     throw custom error with message that file is too large
}

liza-veis (approved):

PR #363, opened 2025-08-13
feat(frontend): toast notification on updating pofile ss-339

timosh-yura:
Added toast notification on updating profile
image


apps/frontend/src/pages/profile/libs/components/user-details/user-details.tsx
```diff
@@ -24,13 +27,24 @@ const UserDetails = (): null | React.JSX.Element => {
...
 		},
 		[dispatch],
 	);
 
+	useEffect(() => {
+		if (isFinished) {
+			toast.success("Profile updated");
+			dispatch(authActions.clearPatchStatus());
+		}
+	}, [isFinished, dispatch]);
```
liza-veis, line 46: We have toastNotifier in the thunk extra arguments, so we can just do it with the patchAuthenticatedUser action creator. Just extract toastNotifier from the extra arguments there and call it after the request awaited

liza-veis (changes requested):

liza-veis (approved):

PR #364, opened 2025-08-13
feat(backend): filter routes by category backend ss-345

DenisDemirov:
Added :

- Migration with creating table Categories and edited route_categories
- Filtration by category feature
- Check categories param in validation schema

Screenshot 2025-08-13 220451_compressed
Screenshot 2025-08-13 172541
зображення

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,6 +1,6 @@
 {
-	".": "1.39.1",
-	"apps/backend": "1.25.2",
+	".": "1.39.2",
+	"apps/backend": "1.25.3",
 	"apps/frontend": "1.31.0",
 	"packages/shared": "1.23.0",
 	"tests": "0.0.9"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [1.39.2](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/compare/v1.39.1...v1.39.2) (2025-08-19)
+
+
+### Bug Fixes
+
+* **backend:** fix routes repository queries ss-432 ([#433](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/issues/433)) ([02a1469](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/commit/02a1469388d9d025e45c5119879140e4890b2279))
+
 ## [1.39.1](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/compare/v1.39.0...v1.39.1) (2025-08-18)
 
 
--- a/apps/backend/.env.example
+++ b/apps/backend/.env.example
@@ -40,4 +40,3 @@ AWS_REGION=eu-north-1
 AWS_S3_BUCKET_NAME=bucket-name
 AWS_MAX_FILE_SIZE_MB=10
 
-
--- a/apps/backend/CHANGELOG.md
+++ b/apps/backend/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [1.25.3](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/compare/@smartscapes/backend-v1.25.2...@smartscapes/backend-v1.25.3) (2025-08-19)
+
+
+### Bug Fixes
+
+* **backend:** fix routes repository queries ss-432 ([#433](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/issues/433)) ([02a1469](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/commit/02a1469388d9d025e45c5119879140e4890b2279))
+
 ## [1.25.2](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/compare/@smartscapes/backend-v1.25.1...@smartscapes/backend-v1.25.2) (2025-08-18)
 
 
--- a/apps/backend/package.json
+++ b/apps/backend/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "@smartscapes/backend",
 	"type": "module",
-	"version": "1.25.2",
+	"version": "1.25.3",
 	"engines": {
 		"node": ">=22.10.0 <23.0.0",
 		"npm": "10.x.x"
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "smartscapes",
-	"version": "1.39.1",
+	"version": "1.39.2",
 	"lockfileVersion": 3,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "smartscapes",
-			"version": "1.39.1",
+			"version": "1.39.2",
 			"workspaces": [
 				"apps/backend",
 				"apps/frontend",
@@ -51,7 +51,7 @@
 		},
 		"apps/backend": {
 			"name": "@smartscapes/backend",
-			"version": "1.25.2",
+			"version": "1.25.3",
 			"dependencies": {
 				"@aws-sdk/client-s3": "^3.0.0",
 				"@fastify/cors": "11.0.1",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "smartscapes",
-	"version": "1.39.1",
+	"version": "1.39.2",
 	"type": "module",
 	"workspaces": [
 		"apps/backend",
--- a/readme.md
+++ b/readme.md
@@ -120,6 +120,7 @@ erDiagram
     decimal distance
     int duration
     geometry geometry
+    int created_by_user_id FK
   }
 
   planned_paths {
--- a/tests/api/controllers/api-controller.ts
+++ b/tests/api/controllers/api-controller.ts
@@ -1,12 +1,15 @@
 import { type APIRequestContext } from "@playwright/test";
 
 import { AuthController } from "./auth-controller.js";
+import { POIController } from "./poi-controler.js";
 
 class ApiController {
 	public readonly auth: AuthController;
+	public readonly poi: POIController;
 
 	public constructor(request: APIRequestContext) {
 		this.auth = new AuthController(request);
+		this.poi = new POIController(request);
 	}
 }
 
--- a/tests/api/controllers/poi-controler.ts
+++ b/tests/api/controllers/poi-controler.ts
@@ -1,0 +1,77 @@
+import { type APIRequestContext, type APIResponse } from "@playwright/test";
+import { type PointsOfInterestRequestDto } from "@smartscapes/shared";
+
+import { APIPath } from "../../consts/api-path.js";
+
+class POIController {
+	public constructor(private request: APIRequestContext) {}
+
+	public async createPoi(
+		poiRequestDto: PointsOfInterestRequestDto,
+		accessToken: string,
+	): Promise<APIResponse> {
+		return await this.request.post(APIPath.POI, {
+			data: poiRequestDto,
+			headers: {
+				Authorization: `Bearer ${accessToken}`,
+			},
+		});
+	}
+
+	public async deletePoi(
+		id: number,
+		accessToken: string,
+	): Promise<APIResponse> {
+		return await this.request.delete(`${APIPath.POI}/${String(id)}`, {
+			headers: {
+				Authorization: `Bearer ${accessToken}`,
+			},
+		});
+	}
+
+	public async getAllPois(
+		accessToken: string,
+		filters?: {
+			latitude?: string;
+			longitude?: string;
+			name?: string;
+			radius?: string;
+		},
+	): Promise<APIResponse> {
+		const query = filters
+			? "?" + new URLSearchParams(filters as Record<string, string>).toString()
+			: "";
+
+		return await this.request.get(`${APIPath.POI}${query}`, {
+			headers: {
+				Authorization: `Bearer ${accessToken}`,
+			},
+		});
+	}
+
+	public async getPoiById(
+		id: number,
+		accessToken: string,
+	): Promise<APIResponse> {
+		return await this.request.get(`${APIPath.POI}/${String(id)}`, {
+			headers: {
+				Authorization: `Bearer ${accessToken}`,
+			},
+		});
+	}
+
+	public async updatePoi(
+		id: number,
+		poiRequestDto: PointsOfInterestRequestDto,
+		accessToken: string,
+	): Promise<APIResponse> {
+		return await this.request.patch(`${APIPath.POI}/${String(id)}`, {
+			data: poiRequestDto,
+			headers: {
+				Authorization: `Bearer ${accessToken}`,
+			},
+		});
+	}
+}
+
+export { POIController };
--- a/tests/api/schemas/poi-schema.ts
+++ b/tests/api/schemas/poi-schema.ts
@@ -1,0 +1,43 @@
+const poiSchema = {
+	"properties": {
+		"data": {
+			"properties": {
+				"description": {
+					"type": "string",
+				},
+				"id": {
+					"type": "integer",
+				},
+				"location": {
+					"properties": {
+						"coordinates": {
+							"items": [
+								{
+									"type": "number",
+								},
+								{
+									"type": "number",
+								},
+							],
+							"type": "array",
+						},
+						"type": {
+							"type": "string",
+						},
+					},
+					"required": ["coordinates", "type"],
+					"type": "object",
+				},
+				"name": {
+					"type": "string",
+				},
+			},
+			"required": ["id", "location", "name", "description"],
+			"type": "object",
+		},
+	},
+	"required": ["data"],
+	"type": "object",
+};
+
+export { poiSchema };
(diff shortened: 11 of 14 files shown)


packages/shared/src/modules/routes/libs/enums/routes-validation-message.enum.ts
```diff
@@ -1,6 +1,10 @@
 import { RoutesValidationRule } from "./routes-validation-rule.enum.js";
 
 const RoutesValidationMessage = {
+	CATEGORY_INVALID_KEY:
+		"Category input key string wrong format, only latin letters without spaces",
```
vldkorin, line 5:
```suggestion 
"Category key must contain only Latin letters without spaces.",

File: packages/shared/src/modules/routes/libs/types/routes-find-all-options.type.ts [hunk 2]
```diff
@@ -1,5 +1,6 @@
 type RoutesFindAllOptions = {
-	name?: string;
+	categories?: string[] | undefined;
+	name?: string | undefined;
```
kalinkaaaa14, line 3: Why do we need to set undefined here? We should either have those query params, or do not have categories/name key query param at all

```suggestion
	categories?: string[];
	name?: string;
```

apps/backend/src/db/migrations/20250813100759_update_route_categories_and_add_categories_table.ts
```diff
@@ -0,0 +1,67 @@
...
+			table
+				.timestamp(ColumnName.CREATED_AT)
+				.notNullable()
+				.defaultTo(trx.fn.now());
+			table
+				.timestamp(ColumnName.UPDATED_AT)
+				.notNullable()
+				.defaultTo(trx.fn.now());
+			table.string(ColumnName.KEY).unique().notNullable();
+			table.string(ColumnName.NAME).notNullable();
```
kalinkaaaa14, line 44: let's add same limit as you have on validation - 32 chars max

apps/backend/src/modules/routes/routes.controller.ts
```diff
@@ -374,9 +380,16 @@ class RoutesController extends BaseController {
 			query?: RoutesFindAllOptions;
 		}>,
 	): Promise<APIHandlerResponse<RoutesResponseDto[]>> {
-		const { query = null } = options;
+		const parsedQuery = routesSearchQueryValidationSchema.parse(
+			options.query ?? {},
+		);
```
kalinkaaaa14, line 385: we can pass categories not like categories=one,second, but like categories=one&categories=second, and on the controller we won't need to make any parsing, we can just use query.categories

packages/shared/src/modules/routes/libs/validation-schemas/routes-search-query.validation-schema.ts
```diff
@@ -5,7 +5,30 @@ import {
 	RoutesValidationRule,
 } from "../enums/enums.js";
 
+const categoryName = z
+	.string()
+	.regex(/^[a-zA-Z-]+$/)
```
liza-veis, line 10: We need to add a message here if the regexp is invalid

packages/shared/src/modules/routes/libs/validation-schemas/routes-search-query.validation-schema.ts
```diff
@@ -5,7 +5,20 @@ import {
 	RoutesValidationRule,
 } from "../enums/enums.js";
 
+const categoryName = z
```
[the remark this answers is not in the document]
DenisDemirov: I do not get what exactly you mean. We do not have logic for category creation, did i need to add it ?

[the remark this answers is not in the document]
liza-veis: We previously had route-categories creation logic, which was intended to be used as categories. But since we move on to the many-to-many relationship and will have route-categories as a transition table, all the logic that we previously had for route-categories, e.g., api for creation and retrieval, should now create and return categories, not route-categories

packages/shared/src/modules/routes/libs/validation-schemas/routes-search-query.validation-schema.ts
```diff
@@ -5,7 +5,20 @@ import {
...
+	.string()
+	.regex(/^[a-zA-Z-]+$/, {
+		message: RoutesValidationMessage.CATEGORY_INVALID_KEY,
+	})
+	.min(RoutesValidationRule.CATEGORY_MINIMUM_LENGTH, {
+		message: RoutesValidationMessage.CATEGORY_MINIMUM_LENGTH,
+	})
+	.max(RoutesValidationRule.CATEGORY_MAXIMUM_LENGTH, {
+		message: RoutesValidationMessage.CATEGORY_MAXIMUM_LENGTH,
+	});
```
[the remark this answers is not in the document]
kalinkaaaa14: It would be better to add validation rule on name field during category creation, I mean to extend existing validation schema

apps/backend/src/db/migrations/20250813100759_update_route_categories_and_add_categories_table.ts
```diff
@@ -0,0 +1,69 @@
...
+	CATEGORY_ID: "category_id",
+	CREATED_AT: "created_at",
+	ID: "id",
+	KEY: "key",
+	NAME: "name",
+	ROUTE_ID: "route_id",
+	UPDATED_AT: "updated_at",
+};
+
+const MAX_NAME_LENGTH = 32;
```
liza-veis, line 19: i would make limit larger
```suggestion
const MAX_NAME_LENGTH = 100;
```

packages/shared/src/modules/routes/libs/validation-schemas/route-search-query.validation-schema.ts
```diff
@@ -5,7 +5,28 @@ import {
...
 const routesSearchQuery = z.object({
+	categories: z
+		.preprocess((value) => {
+			if (value === undefined || value === null) {
+				return;
+			}
+
+			return Array.isArray(value) ? (value as string[]) : [value];
+		}, z.array(categoryKey))
+		.optional(),
```
liza-veis, line 29: take a look at how processing is done for pointsOfInterestSearchQuery. let's transform the value here and remove the same transformation inside the controller
```suggestion
			.transform(parseToFloat)
			.pipe(
```

packages/shared/src/modules/routes/libs/types/types.ts
```diff
@@ -5,3 +5,6 @@ export { type RouteFindAllOptions } from "./route-find-all-options.type.js";
 export { type RouteGetAllItemResponseDto } from "./route-get-all-item-response-dto.type.js";
 export { type RouteGetByIdResponseDto } from "./route-get-by-id-response-dto.type.js";
 export { type RoutePatchRequestDto } from "./route-patch-request.dto.js";
+export { type RouteRequestConstructDto } from "./route-request-construct-dto.type.js";
+export { type RouteRequestCreateDto } from "./route-request-dto.type.js";
+export { type RouteRequestPatchDto } from "./route-request-update.dto.js";
```
liza-veis, line 10:
```suggestion
export { type RouteConstructRequestDto } from "./route-construct-request-dto.type.js";
export { type RouteCreateRequestDto } from "./route-create-request-dto.type.js";
export { type RoutePatchRequestDto } from "./route-patch-request-dto.type.js";
```

packages/shared/src/modules/routes/libs/types/route-response-dto.type.ts
liza-veis, line 1: file name should much the exported type name

apps/backend/src/modules/route-categories/route-category.controller.ts
```diff
@@ -10,15 +10,16 @@ import { type RouteCategoryService } from "~/modules/route-categories/route-cate
```
liza-veis, line 10: Since we are now using this module for categories, not route-categories - we need to rename everything in the module from route categories to categories, including module, service, types, schemas, etc.

liza-veis (changes requested):

packages/shared/src/modules/routes/routes.ts
```diff
@@ -7,6 +7,7 @@ export {
 	type RouteGetAllItemResponseDto,
 	type RouteGetByIdResponseDto,
 	type RoutePatchRequestDto,
+	type RouteRequestCreateDto,
```
kalinkaaaa14, line 10:
```suggestion
	type RouteCreateRequestDto,
```

packages/shared/src/modules/routes/libs/validation-schemas/route-search-query.validation-schema.ts
```diff
@@ -5,7 +5,23 @@ import {
...
+	})
+	.max(RoutesValidationRule.CATEGORY_MAXIMUM_LENGTH, {
+		message: RoutesValidationMessage.CATEGORY_MAXIMUM_LENGTH,
+	});
+
 const routesSearchQuery = z.object({
+	categories: z
+		.union([categoryKey, z.array(categoryKey)])
+		.transform((value) => (Array.isArray(value) ? value : [value]))
+		.optional(),
```
kalinkaaaa14, line 24: it should always be validates as an array in the search
```
 categories: z
    .array(categoryKey)
    .optional(),
```

packages/shared/src/modules/routes/libs/types/types.ts
```diff
@@ -5,3 +5,4 @@ export { type RouteFindAllOptions } from "./route-find-all-options.type.js";
 export { type RouteGetAllItemResponseDto } from "./route-get-all-item-response-dto.type.js";
 export { type RouteGetByIdResponseDto } from "./route-get-by-id-response-dto.type.js";
 export { type RoutePatchRequestDto } from "./route-patch-request.dto.js";
+export { type RouteRequestCreateDto } from "./route-request-dto.type.js";
```
kalinkaaaa14, line 8:
```suggestion
export { type RouteCreateRequestDto } from "./route-create-request-dto.type.js";
```

packages/shared/src/modules/routes/libs/types/route-response-dto.type.ts
kalinkaaaa14, line 1: filename - route-get-by-id-response-dto.type.ts

packages/shared/src/modules/routes/libs/types/route-request-update.dto.ts
```diff
@@ -0,0 +1,6 @@
+type RouteRequestPatchDto = {
```
kalinkaaaa14, line 1: Merge please the latest changes from the main because a lot of same renamings were already done and should be present in main

apps/backend/src/modules/routes/route.service.ts
```diff
@@ -96,12 +96,16 @@ class RouteService implements Service {
 	public async findAll(
 		options: null | RouteFindAllOptions,
 	): Promise<CollectionResult<RouteGetAllItemResponseDto>> {
+		if (options?.categories) {
+			options.categories = Array.isArray(options.categories)
+				? options.categories
+				: [options.categories];
+		}
+
```
kalinkaaaa14, line 104: categories should be always of array type

apps/backend/src/modules/routes/route.repository.ts
```diff
@@ -31,22 +31,28 @@ class RouteRepository implements Repository {
...
-					builder.whereILike("name", `%${options.name.trim()}%`);
-				}
-			});
-
-		return routes.map((point) => RouteEntity.initializeList(point));
+			.select("routes.id", "routes.name", "routes.description");
+
+		if (options?.name?.trim()) {
+			query.whereILike("routes.name", `%${options.name.trim()}%`);
+		}
```
kalinkaaaa14, line 44:
```suggestion
		if (options?.name) {
			query.whereILike("routes.name", `%${options.name.trim()}%`);
		}
```

packages/shared/src/modules/categories/libs/validation-schemas/category-create.validation-schema.ts
```diff
@@ -0,0 +1,33 @@
...
+		.regex(/^[a-zA-Z-]+$/, {
+			message: CatigoriesValidationMessage.CATEGORY_INVALID_KEY,
+		})
+		.min(CategoriesValidationRule.CATEGORY_MINIMUM_LENGTH, {
+			message: CatigoriesValidationMessage.CATEGORY_MINIMUM_LENGTH,
+		})
+		.max(CategoriesValidationRule.CATEGORY_MAXIMUM_LENGTH, {
+			message: CatigoriesValidationMessage.CATEGORY_MAXIMUM_LENGTH,
+		}),
+});
```
kalinkaaaa14, line 31:
```suggestion
const categoryCreate = z.object({
	name: z
		.string()
		.regex(/^[a-zA-Z-]+$/, {
			message: CatigoriesValidationMessage.CATEGORY_INVALID_KEY,
		})
		.min(CategoriesValidationRule.CATEGORY_MINIMUM_LENGTH, {
			message: CatigoriesValidationMessage.CATEGORY_MINIMUM_LENGTH,
		})
		.max(CategoriesValidationRule.CATEGORY_MAXIMUM_LENGTH, {
			message: CatigoriesValidationMessage.CATEGORY_MAXIMUM_LENGTH,
		}),
});
```

packages/shared/src/modules/categories/libs/validation-schemas/category-create.validation-schema.ts
```diff
@@ -0,0 +1,33 @@
+import { z } from "zod";
+
+import {
+	CategoriesValidationRule,
+	CatigoriesValidationMessage,
+} from "../enums/enums.js";
+
+const categoryCreate = z.object({
```
kalinkaaaa14, line 8: The admin will enter only the name when will try to create a new category, key should be autogenerated on the backend service

apps/backend/src/modules/categories/category.service.ts
```diff
@@ -0,0 +1,66 @@
...
+	private categoryRepository: CategoryRepository;
+
+	public constructor(categoryRepository: CategoryRepository) {
+		this.categoryRepository = categoryRepository;
+	}
+
+	public async create(
+		payload: CategoryCreateRequestDto,
+	): Promise<CategoryGetAllItemResponseDto> {
+		const { key, name } = payload;
```
kalinkaaaa14, line 23: You should create helper function - changeToCamelCase, which will be used here.

const key = changeToCamelCase(name);

example: name = "Category Cafe" => key "categoryCafe"

packages/shared/src/modules/routes/libs/validation-schemas/route-search-query.validation-schema.ts
DenisDemirov: When on request we have only one category it interpenetrating on back just as string, not like array<string> with single element. It cause a problem with types
зображення

apps/backend/src/modules/routes/route.repository.ts
liza-veis: @DenisDemirov, it feels to me that it should be trimmed on the validation schema level. Please check this, and maybe this trim is unnecessary here 🤔

packages/shared/src/modules/routes/libs/validation-schemas/route-search-query.validation-schema.ts
liza-veis: @kalinkaaaa14, I think we need a transformation here, for the reason here https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/364#discussion_r2280609410

apps/backend/src/modules/routes/route.repository.ts
DenisDemirov: Yes, i will remove it. It was writed for test before than validation was written - i forgot to remove it

packages/shared/src/libs/helpers/change-string-case/change-string-case.helper.ts
```diff
@@ -0,0 +1,15 @@
+import * as changeCase from "change-case";
+
+import { StringCaseType } from "../../enums/enums.js";
+import { type ValueOf } from "../../types/value-of.type.js";
```
kalinkaaaa14, line 4:
```suggestion
import { type ValueOf } from "../../types/types.js";
```

packages/shared/src/libs/helpers/change-string-case/change-string-case.helper.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const changeStringCase = (
+	string_: string,
+	caseType: ValueOf<typeof StringCaseType>,
+): string => {
+	return caseType === StringCaseType.CAMEL
```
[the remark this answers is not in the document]
liza-veis: resolving this issue after the internal discussion

packages/shared/src/libs/enums/string-case-type.enum.ts
```diff
@@ -0,0 +1,13 @@
+const StringCaseType = {
+	CAMEL: "camelCase",
+	CAPITAL: "capitalCase",
+	CONSTANT: "constantCase",
+	KEBAB: "kebabCase",
+	PASCAL: "pascalCase",
+	PASCAL_SNAKE: "pascalSnakeCase",
+	PATH: "pathCase",
+	SNAKE: "snakeCase",
+	TRAIN: "trainCase",
```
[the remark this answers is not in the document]
liza-veis: Yes, agree, let's use camel case

PR #383, opened 2025-08-14
feat(frontend): cover dashboard items with permissions ss-336

nazartsya:

- added an ability to configure permissions for routes;
- added an ability to configure permissions for sidebar items;
- gave Manage routes page only for users who have manage_routes permission;
- added general logic of redirecting the user to the first page available in the sidebar after the login;
- if a sidebar section has no items, hide the section title (Administration) as well.


apps/frontend/src/libs/helpers/get-first-navigation-item.helper.ts
```diff
@@ -0,0 +1,24 @@
+import {
+	type NavigationItem,
+	type NavigationItemsGroup,
+} from "~/libs/types/types.js";
+import { type PermissionItemDto } from "~/modules/auth/libs/types/types.js";
+
+import { checkPermission } from "./helpers.js";
+
+const FIRST_ITEM = 0;
```
vldkorin, line 9:
```suggestion

```
The rules for the linter have been changed, so now we can use 0 without the “magic numbers” error.

timosh-yura:
lgtm

apps/frontend/src/libs/helpers/check-permission.helper.ts
```diff
@@ -0,0 +1,16 @@
+import { type PermissionItemDto } from "~/modules/users/libs/types/types.js";
+
+const checkPermission = (
```
kalinkaaaa14, line 3:
```suggestion
const checkHasPermission = (
```

apps/frontend/src/libs/helpers/get-permitted-navigation-items.helper.ts
```diff
@@ -0,0 +1,20 @@
...
+	userPermissions: PermissionItemDto[],
+): NavigationItemsGroup[] => {
+	return navigationGroups
+		.map((group) => ({
+			...group,
+			items: group.items.filter((item) =>
+				checkPermission(item.pagePermissions, userPermissions),
+			),
+		}))
+		.filter((group) => group.items.length > 0);
```
kalinkaaaa14, line 17: You can use .reduce here instead of map + filter 

```
return navigationGroups.reduce<Type it>((acc, group) => {
	const filteredItems = group.items.filter((item) =>
		checkHasPermission(item.pagePermissions, userPermissions),
	);

	if (filteredItems.length > 0) {
		acc.push({
			...group,
			items: filteredItems,
		});
	}

	return acc;
}, []);
```

nazartsya: I got a linter error "`Array#reduce()` is not allowed. Prefer other types of loop for readability."

apps/frontend/src/libs/helpers/get-permitted-navigation-items.helper.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { GroupKey } from "../enums/enums.js";
+import { checkHasPermission } from "./helpers.js";
+
+const getPermittedNavigationItems = (
+	navigationGroups: NavigationItemsGroup[],
+	userGroup: GroupResponseDto | null,
+	userPermissions: PermissionItemDto[],
+): NavigationItemsGroup[] => {
+	return navigationGroups
```
h4vefundieyoung, line 15:
```
const filteredGroups = [];
for(const group of navigationGroups){
  const items = group.items.filter((item) => checkHasPermission(group.pagePermissions, userPermissions));
  const isMatchingGroup = userGroup?.key === GroupKey.USERS && group === navigationGroups[0];
  !isMatchingGroup && filteredGroups.push({ ...group, items })
}
```

i'm not really sure what that index check stands for, and i believe you should refactor it a little bit, but this way it would be a little bit more efficient.

h4vefundieyoung, line 15:
```
const filteredGroups = [];
for(const group of navigationGroups){
  const items = group.items.filter((item) => checkHasPermission(item.pagePermissions, userPermissions));
  const isMatchingGroup = !(userGroup?.key === GroupKey.USERS && group === navigationGroups[0]);
  isMatchingGroup && filteredGroups.push({ ...group, items })
}
```

i'm not really sure what that index check stands for, and i believe you should refactor it a little bit, but this way it would be a little bit more efficient.

nazartsya: @liza-veis but if I add pagePermissions to all items in Administration section, the section Administration the section will disappear, maybe that's what was meant

nazartsya: @liza-veis it is logical to assume that all elements in the section admin sidebar should have pagePermissions

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -59,7 +60,19 @@ const Auth = (): React.JSX.Element => {
...
+				[PermissionKey.MANAGE_ROUTES],
+				authenticatedUser.group.permissions,
+			);
+
+		return (
+			<Navigate
+				replace
+				to={isAdminWithManageRoutes ? AppRoute.ROUTES : AppRoute.APP}
+			/>
+		);
```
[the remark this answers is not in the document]
nazartsya: @liza-veis now I have permittedNavigationItems inside Dashboard component. Should I use useEffect in Dashboard for redirection?

apps/frontend/src/libs/helpers/get-permitted-navigation-items.helper.ts
liza-veis: Yes, they will. We just need to render the Administration section if there is at least some permitted route there. So we can simplify this logic by just filtering out sections that don't have permitted routes. I believe I already saw this logic in PR. We won't specifically attach any functionality to the role keys on the frontend or backend. Those roles are used to group permissions and simplify granting / pulling those permissions quickly. In the app logic we need to check everything by permissions, not by the group key.

apps/frontend/src/pages/auth/auth.tsx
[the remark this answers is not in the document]
liza-veis: What we are trying to achieve is that when the user initially gets to the page /app route, to redirect them to the appropriate page. Either Explore for unauthorized, or the first navigation item available for the users. It may look like this in the app component.

```tsx
               //We need to retrieve permissions the same way we do for sidebar, this will need to be passed to the common layout in #130 to avoid repetition with the sidebar
		const userPermissions = authenticatedUser.group.permissions ?? [];
		const permittedNavigationItems = getPermittedNavigationItems(
			NAVIGATION_ITEMS_GROUPS,
			userPermissions
		);
		
useEffect(() => {
		if (!isInitialized || location.pathname !== AppRoute.ROOT) { // we only want redirection for the root page
		   return; 
		}

		// if authenticated user is not present, return explore
		if (!authenticatedUser) {
			navigate(AppRoute.EXPLORE, { replace: true });
			return;
		}

               
		const [navigationGroup] = permittedNavigationItems;
                 const [firstRoute] = navigationGroup ?? [];

		// If nothing was found or this matches root, do nothing
		if (!firstRoute || firstRoute === AppRoute.ROOT) {
                    return;
                 }
                  
                 // otherwise navigate to the first route 
		navigate(firstRoute, { replace: true });
	}, [isInitialized]); // only trigger on initialized
```

kalinkaaaa14 (approved):

liza-veis (approved):

PR #395, opened 2025-08-16
feat(frontend): implement explore routes block ss-270

my-derkach:
Implement explore routes block, namely implement actions, api, thunk for getById in poi and getAll in routes. Create a component explore routes block that create infinite scroll sidebar with routes sorted by the closest starting point to user.

### DIFF
--- a/apps/frontend/src/libs/components/explore-routes-block/constants/plane-angel.constant.ts
+++ b/apps/frontend/src/libs/components/explore-routes-block/constants/plane-angel.constant.ts
@@ -1,0 +1,3 @@
+const PLANE_ANGEL = 180;
+
+export { PLANE_ANGEL };
--- a/apps/frontend/src/libs/components/explore-routes-block/explore-routes-block.tsx
+++ b/apps/frontend/src/libs/components/explore-routes-block/explore-routes-block.tsx
@@ -1,0 +1,67 @@
+import React from "react";
+
+import { useAppDispatch, useEffect, useState } from "~/libs/hooks/hooks.js";
+import { type RouteGetByIdResponseDto } from "~/modules/routes/libs/types/types.js";
+
+import { Loader, RouteCard } from "../components.js";
+import { getCurrentUserPosition, getSortedRoutes } from "./helpers/helpers.js";
+import styles from "./styles.module.css";
+
+const ExploreRoutesBlock = (): React.JSX.Element => {
+	const dispatch = useAppDispatch();
+	const [routes, setRoutes] = useState<RouteGetByIdResponseDto[]>([]);
+	const [loading, setLoading] = useState<boolean>(true);
+	const [error, setError] = useState<null | string>(null);
+
+	useEffect(() => {
+		const fetchData = async (): Promise<void> => {
+			try {
+				const userCoordinates = await getCurrentUserPosition();
+				const sortedRoutes = await getSortedRoutes(dispatch, userCoordinates);
+				setRoutes(sortedRoutes);
+			} catch (error: unknown) {
+				if (error instanceof Error) {
+					setError(error.message);
+				} else {
+					setError("An unknown error occured.");
+				}
+			} finally {
+				setLoading(false);
+			}
+		};
+
+		void fetchData();
+	}, [dispatch]);
+
+	let content;
+
+	if (loading) {
+		content = (
+			<div className={styles["loading"]}>
+				<Loader />
+			</div>
+		);
+	} else if (error) {
+		content = <div className={styles["error"]}>Error: {error}</div>;
+	} else {
+		content =
+			routes.length === 0 ? (
+				<div className={styles["not-found"]}>No routes found nearby.</div>
+			) : (
+				<ul className={styles["list"]}>
+					{routes.map((route) => (
+						<RouteCard imageUrl={null} key={route.id} name={route.name} />
+					))}
+				</ul>
+			);
+	}
+
+	return (
+		<div className={styles["container"]}>
+			<h3 className={styles["title"]}>Explore routes</h3>
+			{content}
+		</div>
+	);
+};
+
+export { ExploreRoutesBlock };
--- a/apps/frontend/src/libs/components/explore-routes-block/helpers/get-current-user-position.helper.ts
+++ b/apps/frontend/src/libs/components/explore-routes-block/helpers/get-current-user-position.helper.ts
@@ -1,0 +1,15 @@
+const getCurrentUserPosition = (): Promise<[number, number]> => {
+	return new Promise((resolve, reject) => {
+		// eslint-disable-next-line sonarjs/no-intrusive-permissions
+		navigator.geolocation.getCurrentPosition(
+			(position) => {
+				resolve([position.coords.latitude, position.coords.longitude]);
+			},
+			(error) => {
+				reject(new Error(`Failed to get position: ${error.message}`));
+			},
+		);
+	});
+};
+
+export { getCurrentUserPosition };
--- a/apps/frontend/src/libs/components/explore-routes-block/helpers/get-sorted-routes.helper.ts
+++ b/apps/frontend/src/libs/components/explore-routes-block/helpers/get-sorted-routes.helper.ts
@@ -1,0 +1,34 @@
+import { type useAppDispatch } from "~/libs/hooks/hooks.js";
+import { type RouteGetByIdResponseDto } from "~/modules/routes/libs/types/types.js";
+
+import { getDistance, getPOICoordinates, getRoutes } from "./helpers.js";
+
+const getSortedRoutes = async (
+	dispatch: ReturnType<typeof useAppDispatch>,
+	userCoordinates: [number, number],
+): Promise<RouteGetByIdResponseDto[]> => {
+	const allRoutes = await getRoutes(dispatch);
+	const routesWithDistances = await Promise.all(
+		allRoutes.map(async (route) => {
+			const startingPoi = route.pois.find((poi) => poi.visitOrder === 0);
+
+			if (!startingPoi) {
+				return { ...route, distanceToUser: Infinity };
+			}
+
+			const pointCoordinates = await getPOICoordinates(
+				dispatch,
+				startingPoi.id,
+			);
+			const distance = getDistance(userCoordinates, pointCoordinates);
+
+			return { ...route, distanceToUser: distance };
+		}),
+	);
+
+	return routesWithDistances.sort(
+		(a, b) => a.distanceToUser - b.distanceToUser,
+	);
+};
+
+export { getSortedRoutes };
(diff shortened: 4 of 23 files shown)


apps/frontend/src/libs/components/explore-routes-block/explore-routes-block.tsx
```diff
@@ -0,0 +1,67 @@
...
+import styles from "./styles.module.css";
+
+const ExploreRoutesBlock = (): React.JSX.Element => {
+	const dispatch = useAppDispatch();
+	const [routes, setRoutes] = useState<RouteGetByIdResponseDto[]>([]);
+	const [loading, setLoading] = useState<boolean>(true);
+	const [error, setError] = useState<null | string>(null);
+
+	useEffect(() => {
+		const fetchData = async (): Promise<void> => {
```
h4vefundieyoung, line 17: well, i've already typed it in your draft pr, not sure u've seen it.
i believe we can't have such async logic in the component.
[check it here](https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/337#discussion_r2277655384)

apps/frontend/src/modules/points-of-interest/points-of-interest-api.ts
```diff
@@ -36,6 +36,23 @@ class PointOfInterestApi extends BaseHTTPApi {
...
 	}
+
+	public async getById(
+		id: number,
+	): Promise<APIResponse<PointsOfInterestResponseDto>> {
+		const response = await this.load<APIResponse<PointsOfInterestResponseDto>>(
+			this.getFullEndpoint(
+				`${PointsOfInterestApiPath.ROOT}${id.toString()}`,
+				{},
+			),
```
mxm-kotyk, line 47:
```suggestion
			this.getFullEndpoint(
				PointsOfInterestApiPath.ID,
				{id: id.toString()},
			),
```
And for the `PointsOfInterestApiPath`:
```
const PointsOfInterestApiPath = {
  ID: "/:id",
	ROOT: "/",
} as const;
```

apps/frontend/src/libs/components/explore-routes-block/constants/plane-angel.constant.ts
```diff
@@ -0,0 +1,3 @@
+const PLANE_ANGEL = 180;
```
mxm-kotyk, line 1:
```suggestion
const PLANE_ANGLE = 180;
```

apps/frontend/src/modules/routes/routes-api.ts
```diff
@@ -19,6 +23,27 @@ class RoutesApi extends BaseHTTPApi {
...
+		query?: RouteFindAllOptions,
+	): Promise<APIResponse<RouteGetByIdResponseDto[]>> {
+		const queryParameters: Record<string, string> = {};
+
+		if (query?.name) {
+			queryParameters["name"] = query.name;
+		}
+
+		const response = await this.load<APIResponse<RouteGetByIdResponseDto[]>>(
+			this.getFullEndpoint(RoutesApiPath.ROOT, queryParameters),
```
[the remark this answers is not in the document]
liza-veis: agree

apps/frontend/src/libs/components/explore-routes-block/explore-routes-block.tsx
liza-veis: yes, for getCurrentPosition logic, we need to creat app slice (for common app actions) and move all the logic to action creator. For the getSortedRoutes we need a separate slice for this page and handle all the logic through it. Any errors should also be set using redux

apps/frontend/src/libs/components/explore-routes-block/helpers/get-sorted-routes.helper.ts
```diff
@@ -0,0 +1,34 @@
...
+			);
+			const distance = getDistance(userCoordinates, pointCoordinates);
+
+			return { ...route, distanceToUser: distance };
+		}),
+	);
+
+	return routesWithDistances.sort(
+		(a, b) => a.distanceToUser - b.distanceToUser,
+	);
```
liza-veis, line 31: All this should be done on the backend in one query; on the frontend, we need to follow the usual approach - just dispatch the getRoutes action and pass user coordinates to it, similarly to how it's done for finding closest POIs on the backend. The query should look the same, with the only difference that we start with selecting routes, then join POI included in route with visit_order one, and then apply the sorting the same way it's done in search closest POI to the coordinates. We don't need all these helpers, distance calculation, and dispatch unwraps. 

When the user goes to the explore page, dispatch the action to get the current user location. Use those coordinates in query params to send a request to the backend using the usual redux flow we have in the app. On the backend, the logic should be very simple too - without any JS helpers. We already have everything needed on the DB level. If you need any help or guidance with implementing this, I believe @vyacheslavKhoroshev could help.

liza-veis (changes requested):

apps/frontend/src/libs/components/explore-routes-block/helpers/get-current-user-position.helper.ts
```diff
@@ -0,0 +1,15 @@
+const getCurrentUserPosition = (): Promise<[number, number]> => {
+	return new Promise((resolve, reject) => {
+		// eslint-disable-next-line sonarjs/no-intrusive-permissions
```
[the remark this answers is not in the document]
my-derkach: Do not yet, but thanks, should to try)

apps/frontend/src/modules/routes/routes-api.ts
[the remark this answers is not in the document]
my-derkach: Yeah, in my previous closed pr I have an implementation of getFullEndpoint that works with different type of query (array also), but since i need only names I make some simpler thing. But I completely agree we need to make some general implementation for future scalability

apps/frontend/src/libs/components/explore-routes-block/helpers/get-sorted-routes.helper.ts
my-derkach: To be honest, I do not fully understand what i need to do. So what comes to my mind first of all as an idea that I need maybe use already implemented find all in routes (maybe even better will be create another endpoint) with query parameters of user current location and inside the repository for this endpoint make a query to database that will select all routes, calculate distance of the starting point and order it. Is it a way?

liza-veis: Take a look at PointsOfInterest module's findAll implementation. We basically need the same, with the only difference on the repository level in how we make a query. We need to pass longitude and latitude as optional parameters, the same way we did it for POIs findAll, and we need to have the same logic in the validation schema, where we check that either both lat/lon are present or none. So, in short, everything is fully the same; the only difference is on the repository level. For pois, we were ordering them just by nearest distance, in the findNearby method. While for routes, we will need to first join the POI with visiting order 0 and then order by distance to it. But it would still look pretty much the same, just with additional join line

apps/frontend/src/libs/modules/api/base-http-api.ts
```diff
@@ -69,15 +68,34 @@ class BaseHTTPApi implements HTTPApi {
...
 
-		return configureString(
-			this.baseUrl,
-			this.path,
-			...(copiedParameters as string[]),
-			options,
-		);
+		let path = this.constructPath(copiedParameters as string[]);
+
+		path = this.replaceDynamicPathSegments(path, options);
```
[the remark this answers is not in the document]
my-derkach: Yeah, agree, my fault, I just tried to implement to get correct url with query, but got problems with configureString. I will update rn

apps/backend/src/modules/routes/route.repository.ts
```diff
@@ -72,12 +75,29 @@ class RouteRepository implements Repository {
...
-					builder.whereILike("name", `%${options.name.trim()}%`);
-				}
-			});
+			]);
+
+		if (name) {
+			query.whereILike("routes.name", `%${name.trim()}%`);
+		}
+
+		if (latitude !== undefined && longitude !== undefined) {
```
lavandosya, line 84: nitpick:
```suggestion
		if (latitude && longitude) {
```

apps/frontend/src/libs/helpers/get-url-with-query-string.helper.ts
```diff
@@ -0,0 +1,26 @@
...
+	if (!queryString) {
+		return url;
+	}
+
+	const parameters = new URLSearchParams();
+
+	for (const [key, value] of Object.entries(queryString)) {
+		if (Array.isArray(value)) {
+			for (const v of value) {
+				parameters.append(key, String(v));
```
lavandosya, line 14:
```suggestion
		if (Array.isArray(value)) {
        const valueArray = value;
        for (const item of valueArray) {
            parameters.append(key, String(item));
        }
```

my-derkach:
@liza-veis I have a question regarding this comment: https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/395/#discussion_r2284676820

I think that we need to have bigger paddings and different for each side, because for unauthorized user we have header, so we need to take into account it height also. And for authorized user we have sidebar, so I think that for padding we also need bigger padding from left or maybe make dynamic padding, for example when we have user we will extend this padding. What approach we will use

apps/frontend/src/modules/routes/routes-api.ts
```diff
@@ -19,6 +24,23 @@ class RoutesApi extends BaseHTTPApi {
 		super({ baseUrl, http, path: APIPath.ROUTES, storage });
 	}
 
+	public async getAll(
+		query?: RouteFindAllOptions,
+	): Promise<APIResponse<RouteGetByIdResponseDto[]>> {
+		const url = getUrlWithQueryString(RoutesApiPath.ROOT, query);
```
liza-veis, line 30: pass the query itself to the load method options as a plain object

apps/frontend/src/pages/explore/libs/components/routes-panel/styles.module.css
```diff
@@ -0,0 +1,29 @@
+.container {
+	display: flex;
+	flex-direction: column;
+	gap: 24px;
+	max-height: calc(100vh - 110px);
```
liza-veis, line 5: It should take 100% height of its parent; it should not depend on the viewport. And the parent container (that has absolute positioning) should also take 100% of the height, but with a margin: 32px. Pay attention to this comment https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/395/files#r2284676820

liza-veis (changes requested):

apps/frontend/src/modules/routes/routes-api.ts
liza-veis: yes, you need to use getUrlWithQueryString inside it and add posibility to pass query there

PR #396, opened 2025-08-16
feat(frontend): add list to dropdown menu ss-338

my-derkach:
Create customizable list for dropdown menu and style it
image


apps/frontend/src/libs/components/link/styles.module.css
```diff
@@ -1,5 +1,6 @@
 .link {
 	color: var(--color-text-brand);
+	text-decoration: none;
 	transition: all 0.3s ease;
```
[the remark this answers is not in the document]
liza-veis: i agree, we want to have text decoration for links, so this should be removed

apps/frontend/src/libs/constants/dropdown-menu-items.constnant.ts
```diff
@@ -0,0 +1,9 @@
+import { AppRoute } from "~/libs/enums/enums.js";
+
+import { type DropdownMenuItem } from "../types/types.js";
+
+const DROPDOWN_MENU_ITEMS: DropdownMenuItem[] = [
+	{ href: AppRoute.PROFILE, label: "Go to Profile" },
```
liza-veis, line 6: let's just name it profile as in sidebar
```suggestion
	{ href: AppRoute.PROFILE, label: "Profile" },
```

kalinkaaaa14:
Also it looks like the dropdown is too large in comparison with user component in the header, let's make it a bit smaller?

liza-veis:
@my-derkach could you please attach a screenshot of how it looks like right now

my-derkach:
https://github.com/user-attachments/assets/118a386d-c55a-4baf-b056-ecf64a137e3c

my-derkach:
https://github.com/user-attachments/assets/cc0b717b-1012-4788-9a5a-32d2bac4ce45

kalinkaaaa14 (approved):
Looks good to me, please resolve Liza's comments

apps/frontend/src/libs/components/dropdown/dropdown.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { type DropdownItem } from "~/libs/types/types.js";
+
+import styles from "./styles.module.css";
+
+type Properties = {
+	items: DropdownItem[];
+};
+
+const Dropdown = ({ items }: Properties): JSX.Element => {
+	const linkClassName = useCallback(() => styles["link"], []);
```
my-derkach, line 14: I have a question about that part, since we are using NavLink we need to pass a function to the className, is it okay if it will be looking like that. Or maybe it would better to come with basic Link?

apps/frontend/src/libs/components/header/libs/components/authenticated-header/authenticated-header.tsx
```diff
@@ -33,6 +34,10 @@ const AuthenticatedHeader = ({ user }: Properties): JSX.Element => {
 		void dispatch(authActions.logout());
 	}, [dispatch]);
 
+	const items: DropdownItem[] = HEADER_DROPDOWN_ITEMS.map((item) =>
+		item.label === "Log out" ? { ...item, onClick: handleLogout } : item,
+	);
```
liza-veis, line 39: We have two good options here: either use useMemo and define all the options right here, or instead of  HEADER_DROPDOWN_ITEMS we should have a helper like - getHeaderDropdownItems that will accept handleLogout in params. Re-mapping is not intuitive and hard to troubleshoot. For our case, I would proceed with just defining items here, without a separate constant, but you can choose an approach with a helper if you like it more. And don't forget to wrap it in useMemo too

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -63,15 +63,15 @@ const Auth = (): React.JSX.Element => {
 		<main className={styles["container"]}>
 			<div className={styles["left-panel"]}>
 				<div className={styles["logo"]}>
-					<Link to={AppRoute.ROOT}>
+					<TextLink to={AppRoute.ROOT}>
```
liza-veis, line 66: seems like for images and logo it should be a usual link, not a text link, even so it used this component before and the same for other places in this pr. We only want to keep text link for pure text

liza-veis (changes requested):

kalinkaaaa14 (approved):

PR #397, opened 2025-08-16
feat(frontend): implement common layout component ss-130

h4vefundieyoung:
Changes:

- Common layout component has been added
- Sign-up, sign-in pages have been excluded from the layout rendering list.

https://github.com/user-attachments/assets/ae202796-3035-4e5c-a1de-b7ba2becf0ee


h4vefundieyoung:
> > i believe we should also exclude layout for the profile page coz it looks kinda strange rn. The other thing is siderbar - probably we should extend it's logic to hide modal after clicking the button. And yea, seems like there is also bug with multiple active tabs when we switching.
> 
> In this layout component, we need to add a common container that will take the whole width and height and will account for sidebar sizes. And we need to go through the rest of the pages and remove unnecessary containers or styles that we had for this. Not the content container, but so that every page looks ok with the sidebar. Yes, after switching pages, we need to hide the sidebar. Let me know if it's better to be done as a separate ticket. For multiple active items - let's ask qa to create a ticket

I've already removed everything in this pr. The case is sidebar - I've already described a few things, and one more is positioning. It bounded to the viewport with fixed positioning. There is nothing I can handle with wrapper without changing sidebar itself. I believe it requires some time to refactor all these sidebar stuff, and it's better to do it in the separate ticket.

liza-veis:
> I've already removed everything in this pr. The case is sidebar - I've already described a few things, and one more is positioning. It bounded to the viewport with fixed positioning. There is nothing I can handle with wrapper without changing sidebar itself. I believe it requires some time to refactor all these sidebar stuff, and it's better to do it in the separate ticket.

Hm, can't we just allocate space for the sidebar in the wrapper? Or how do we handle this on the dashboard page? I believe we have a fixed width for the sidebar, so we just need to move it to the variables and allocate this space in the wrapper

h4vefundieyoung:
> > I've already removed everything in this pr. The case is sidebar - I've already described a few things, and one more is positioning. It bounded to the viewport with fixed positioning. There is nothing I can handle with wrapper without changing sidebar itself. I believe it requires some time to refactor all these sidebar stuff, and it's better to do it in the separate ticket.
> 
> 
> 
> Hm, can't we just allocate space for the sidebar in the wrapper? Or how do we handle this on the dashboard page? I believe we have a fixed width for the sidebar, so we just need to move it to the variables and allocate this space in the wrapper

Well, if that's ok to define that space just with a fixed padding then it's definitely can be done

liza-veis:
Yes, I think it's how it's currently done for the dashboard page; it's the only way for us, considering sidebar uses fixed positioning. Just use variables for this

apps/frontend/src/libs/components/main-layout/main-layout.tsx
```diff
@@ -0,0 +1,35 @@
...
+type Properties = PropsWithChildren & {
+	layoutExclude: ValueOf<typeof AppRoute>[];
+};
+
+const MainLayout = ({ children, layoutExclude }: Properties): JSX.Element => {
+	const { pathname } = useLocation();
+	const user = useAppSelector(({ auth }) => auth.authenticatedUser);
+	const isLayoutInclude = !layoutExclude.some((excludedPath) =>
+		pathname.endsWith(excludedPath),
+	);
```
liza-veis, line 20:
```suggestion
const AppLayout = ({ children, excludedRoutes }: Properties): JSX.Element => {
	const { pathname } = useLocation();
	const user = useAppSelector(({ auth }) => auth.authenticatedUser);
	const isRouteExcluded = excludedRoutes.some((excludedPath) =>
		pathname.endsWith(excludedPath),
	);
```

apps/frontend/src/libs/components/main-layout/main-layout.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+	return (
+		<>
+			{isLayoutInclude && (
+				<>
+					<Header actions={[]} user={user} />
+					{user && <Sidebar navigationItemsGroups={NAVIGATION_ITEMS_GROUPS} />}
```
liza-veis, line 27: Let's use boolean vars for conditional rendering. It's maybe not that important for objects, so it's more like a convention here, but it's a must-have for strings
```suggestion
					{hasUser && <Sidebar navigationItemsGroups={NAVIGATION_ITEMS_GROUPS} />}
```

apps/frontend/src/libs/components/main-layout/main-layout.tsx
```diff
@@ -0,0 +1,35 @@
...
+	const user = useAppSelector(({ auth }) => auth.authenticatedUser);
+	const isLayoutInclude = !layoutExclude.some((excludedPath) =>
+		pathname.endsWith(excludedPath),
+	);
+
+	return (
+		<>
+			{isLayoutInclude && (
+				<>
+					<Header actions={[]} user={user} />
```
liza-veis, line 26: Shouldn't the header contain sign-in in e.g., for unauthorized users on the explore page?

apps/frontend/src/libs/components/main-layout/main-layout.tsx
```diff
@@ -0,0 +1,35 @@
+import { type JSX, type PropsWithChildren } from "react";
+import { useLocation } from "react-router";
```
liza-veis, line 2:
```suggestion
import { useLocation } from "~/libs/hooks/hooks.js";
```

liza-veis (changes requested):

Issue #403, opened 2025-08-17
feat: Load and display Routes list on POI details page


liza-veis:
### What feature?

- [x] Load and display scrollable routes list on poi details page
- [x] On item click, the user should be redirected to route details page -`routes/:id`

### Screenshots

<img width="1318" height="494" alt="Image" src="https://github.com/user-attachments/assets/247c0b1e-409c-4fdd-8cac-e463dbc93573" />

liza-veis:
Show work similarly to #390

Anna-Kutsenko:
<img width="1537" height="947" alt="Image" src="https://github.com/user-attachments/assets/6a3f1a7d-44e2-454d-9c2d-3c89eafc5324" />

PR #417, opened 2025-08-18
feat(frontend/backend): load and display pois list on route details page ss-390

nazartsya:

- loaded and displayed scrollable POIs list on route details page
- on item click, the user will be redirected to poi details page


apps/backend/src/modules/routes/route.service.ts
```diff
@@ -77,14 +78,19 @@ class RouteService implements Service {
...
+			pois: payload.poiIds.map((id, index) => {
+				const poi = poisResult.items.find((p) => p.id === id);
+
+				return {
+					id,
+					name: poi?.name ?? "",
+					visitOrder: index,
+				};
+			}),
 		};
```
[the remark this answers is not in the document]
nazartsya: @liza-veis Yes I understand it but in route.entity.ts there is an error in the initializeNew method because name is required (RouteEntity.initializeNew()). And that's why I made name optional the first time. How to be better with the name field? Make name optional again? now I just passed name: ""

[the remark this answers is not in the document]
liza-veis: initialize new should return the object that will be used for creation, that's the only purpose of it's usage, so we need to adjust it return the right format and eliminate name from ti

apps/backend/src/modules/routes/route.service.ts
```diff
@@ -83,6 +83,7 @@ class RouteService implements Service {
 			...payload,
 			pois: payload.poiIds.map((id, index) => ({
 				id,
+				name: "",
```
liza-veis, line 86: We should not expect a name for initializeNew
```suggestion
```

apps/frontend/src/libs/components/point-of-interest-card/point-of-interest-card.tsx
```diff
@@ -0,0 +1,37 @@
...
+	id,
+	imageUrl,
+	name,
+}: Properties): React.JSX.Element => {
+	return (
+		<li className={styles["card"]}>
+			<Link
+				to={configureString(AppRoute.POINTS_OF_INTEREST_$ID, {
+					id: id.toString(),
+				})}
```
liza-veis, line 24: let's move it outside of render to keep it clean

```
const poiDetailsLink = configureString(AppRoute.POINTS_OF_INTEREST_$ID, {
					id: id.toString(),
				});
```

liza-veis (changes requested):
Two comments, everything else looks good

apps/backend/src/modules/routes/route.service.ts
nazartsya: @liza-veis I need poi name inside route.entity.ts but I have error with create (don't need name) I tried different options. If there is a better option please tell me.

apps/backend/src/modules/routes/route.service.ts
```diff
@@ -84,7 +84,7 @@ class RouteService implements Service {
 			pois: payload.poiIds.map((id, index) => ({
 				id,
 				visitOrder: index,
-			})),
+			})) as { id: number; name: string; visitOrder: number }[],
```
liza-veis, line 87: What I meant by the previous comment is that there shouldn't be any mention of the name in initalizeNew for the routeEntity. It means it should expect the interface without name in the params and it should return the interface without a name. Is there an issue with this? 
```suggestion
			}))
```

nazartsya: @liza-veis initializeNew returns RouteEntity and the name field in pois is required and then an error

apps/frontend/src/libs/components/point-of-interest-card/point-of-interest-card.tsx
h4vefundieyoung, line 1: @liza-veis the component basically has been copied from landing/popular-section, and we will have one more such component in #403. makes sense to make some kind of common scrollable list according to the dry principle.

liza-veis: Yes, I agree, but not in a current task already

apps/backend/src/modules/routes/route.service.ts
liza-veis: I think there was a slight misunderstanding. What we need to do is to make it optional on entity level, and eliminate it for initializeNew and toNewObject object methods, while keeping it as required in toListObject and toObject methods (we will need to cast their types for this). What we want to achieve, is just to be able to use initializeNew and toNewObject without name, while all other methods should have name as required. In other places, like model and response dto, we still want to keep name as required

apps/frontend/src/libs/components/point-of-interest-card/point-of-interest-card.tsx
h4vefundieyoung, line 1: I believe it won't be reused throughout the app and it makes sense to move it to the page folder

PR #442, opened 2025-08-19
feat(frontend): add create route modal ss-360

### LINKED ISSUE #360: feat: Add create route modal
### What feature?

 - [ ] Implement a modal to create Route with fields: Name (text), Description (textarea), and Route. Similarly to #244 
 - [ ] For the Route field, just display `Construct route` button under the label and redirect the user to ` `/routes/construct?returnTo=/routes?modal=create-route` onClick.
 - [ ] The modal should be opened with ?modal=create-route query parameter, because we will need to preserve state for it. (We will add the same for the create poi modal later)
 - [ ] After the `/routes/construct` logic is implemented, it will redirect back to the URL `/routes?modal=create-route&plannedRouteId=id`
 - [ ] if plannedRouteId is present in query params, we will consider it existing and let's display some default test like `Route constructed` under the route field (we will adjust it later). This field is required.
- [ ] We also need to save and restore the name and description to LocalStorage, and restore them when we reload the page. (This is needed because we will have a redirection from another page.)
- [ ] When the user clicks create, we need to send `POST /routes` request with name, description, and planned route ID (planned route ID will be added here #349, for now you can just skip this field in the final payload) 

### Screenshots

Image

ZakZh:
This pull request introduces a global modal management system to the frontend application and adds a new `CreateRouteForm` component for route creation, along with improvements to button styling and form handling. 

**Global Modal Architecture Refactor:**

* Introduced a global modal management system via the new `useModal` hook, allowing modals to be controlled through URL query parameters and shared state. The `Modal` component now renders registered modal components with injected lifecycle props, replacing the previous local state approach. 

**New Feature: Create Route Form**

* Added the `CreateRouteForm` component for creating new routes, supporting form validation, persistent draft storage, and integration with the modal system. Includes supporting constants, styles, and storage key. 

**UI/UX Enhancements:**

* Updated the `Button` component to support a new `outlined` variant with corresponding styles for better visual distinction, especially in forms. 

**Form Modal Integration Updates:**

* Refactored `CreatePOIModal` to work with the new modal system, removing local modal state and using injected lifecycle props for closing and cleanup. 

Closes #360


apps/frontend/src/libs/components/button/styles.module.css
```diff
@@ -38,9 +28,51 @@
...
+
+.button-outlined:hover:not(:disabled) {
+	color: var(--color-text-inverse-strong);
+	background-color: var(--color-fill-brand-strong);
+}
+
+.button-outlined:focus-visible:not(:disabled) {
+	color: var(--color-text-inverse-strong);
+	background-color: var(--color-fill-brand-strong);
 }
```
mxm-kotyk, line 76:
```suggestion
.button-outlined:hover:not(:disabled),
.button-outlined:focus-visible:not(:disabled) {
	color: var(--color-text-inverse-strong);
	background-color: var(--color-fill-brand-strong);
}
```

apps/frontend/src/libs/components/create-route-form/create-route-form.tsx
```diff
@@ -0,0 +1,203 @@
...
+						<span className={styles["label"]}>Route</span>
+						{isRouteConstructed && (
+							<div>
+								<span>Route constructed</span>
+							</div>
+						)}
+						<div>
+							<Button
+								label="Construct"
+								to="/routes/construct?returnTo=/routes?modal=create-route"
```
mxm-kotyk, line 175: I have a feeling that this should be a template string with enums. Something like this maybe:

```suggestion
	to=`${AppRoute.ROUTES_CONSTRUCT}?returnTo=${AppRoute.ROUTES}?modal=${ModalType.CREATE_ROUTE}"
```
But I'm not completely sure if we need this here. And, also there is this "/app" in these enums. What do you think?

apps/frontend/src/libs/components/create-poi-modal/create-poi-modal.tsx
```diff
@@ -13,20 +15,25 @@ import {
...
 	defaultLatitude,
 	defaultLongitude,
-	isOpen,
-	onClose,
 	onSubmit,
 }: Properties): React.JSX.Element => {
+	const createStatus = useAppSelector(
+		(state) => state.pointsOfInterest.createStatus,
+	);
+	const previousCreateStatusReference = useRef<ValueOf<typeof DataStatus>>(
```
h4vefundieyoung, line 33: just wondering what's the purpose of this?

kalinkaaaa14:
Administrative page is already merged, let's add route creation there?

liza-veis (changes requested):

apps/frontend/src/libs/components/create-route-form/create-route-form.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+const CreateRouteForm = ({
+	createRouteFormData,
+	onFormChange,
+	onSubmit,
+}: Properties): React.JSX.Element => {
+	const { control, errors, handleSubmit, handleValueSet } =
+		useAppForm<RouteCreateRequestDto>({
+			defaultValues: DEFAULT_CREATE_ROUTE_PAYLOAD,
+			// validationSchema: routesCreateValidationSchema,
```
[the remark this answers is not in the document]
ZakZh: Uncommented it, but right now the form is always considered invalid, so submitting does nothing and there's no feedback (while there are no errors for POIs and plannedRouteId)

kalinkaaaa14 (approved):

apps/frontend/src/libs/components/create-route-form/create-route-form.tsx
```diff
@@ -0,0 +1,115 @@
...
+	const [name, description, plannedPathId] = useWatch({
+		control,
+		name: ["name", "description", "plannedPathId"],
+	});
+
+	useEffect(() => {
+		if (onFormChange) {
+			onFormChange({ description, name });
+		}
+	}, [description, name, onFormChange]);
```
liza-veis, line 50: We don't need to update the store on form change; we only need to use the state as the initial value (we set it from local storage once, and that's all). On an unload or visibility change event, we need to retrieve values there. Instead of useWatch, let's use getValues from react-hook-form and only call it once when we save data to local storage. It seems like for this we will either need to move useAppForm outside of create-route-form, or move the event handler here inside
```suggestion
```

PR #454, opened 2025-08-20
feat(frontend): show routes without geolocation ss-453

my-derkach:
Implement possibility to display all routes, even if user deny its geolocation
image

### DIFF
--- a/apps/frontend/src/pages/explore/libs/components/routes-panel/routes-panel.tsx
+++ b/apps/frontend/src/pages/explore/libs/components/routes-panel/routes-panel.tsx
@@ -43,6 +43,8 @@ const RoutesPanel = (): React.JSX.Element => {
 		}
 	}, [locationDataStatus, location, dispatch]);
 
+	const hasLocationError = Boolean(locationError);
+
 	const content = useMemo(() => {
 		if (dataStatus === DataStatus.PENDING) {
 			return <Loader />;
@@ -60,7 +62,7 @@ const RoutesPanel = (): React.JSX.Element => {
 
 		return (
 			<>
-				{locationError?.trim() && (
+				{hasLocationError && (
 					<div className={styles["warning"]}>
 						<div>Location access failed: {locationError}.</div>
 						<div>Please enable geolocation to see routes near you.</div>
@@ -74,7 +76,7 @@ const RoutesPanel = (): React.JSX.Element => {
 				</ul>
 			</>
 		);
-	}, [dataStatus, error, locationError, routes]);
+	}, [dataStatus, error, locationError, hasLocationError, routes]);
 
 	return (
 		<div className={styles["container"]}>


apps/frontend/src/pages/explore/libs/components/routes-panel/routes-panel.tsx
```diff
@@ -56,6 +60,12 @@ const RoutesPanel = (): React.JSX.Element => {
 
 		return (
 			<ul className={styles["list"]}>
+				{locationError && (
```
liza-veis, line 63: avoid using non-bolean vars for conditional rendering, especially for strings. If locationError will be an empty string, it will return it and render as unnecessary element

apps/frontend/src/pages/explore/libs/components/routes-panel/routes-panel.tsx
```diff
@@ -44,8 +48,8 @@ const RoutesPanel = (): React.JSX.Element => {
 			return <Loader />;
 		}
 
-		if (error || locationError) {
-			return <span className={styles["error"]}> {error ?? locationError}</span>;
+		if (error) {
+			return <span className={styles["error"]}> {error}</span>;
```
liza-veis, line 52:
```suggestion
			return <span className={styles["error"]}>{error}</span>;
```

apps/frontend/src/pages/explore/libs/components/routes-panel/routes-panel.tsx
```diff
@@ -55,11 +59,20 @@ const RoutesPanel = (): React.JSX.Element => {
...
-				{routes.map((route) => (
-					<RouteCard imageUrl={null} key={route.id} name={route.name} />
-				))}
-			</ul>
+			<>
+				{locationError?.trim() && (
+					<div className={styles["warning"]}>
+						Location access failed: {locationError}. <br />
+						Please enable geolocation to see routes near you.
+					</div>
```
kalinkaaaa14, line 67: Try to get rid of <br/> tag

apps/frontend/src/pages/explore/libs/components/routes-panel/routes-panel.tsx
```diff
@@ -55,11 +59,20 @@ const RoutesPanel = (): React.JSX.Element => {
 		}
 
 		return (
-			<ul className={styles["list"]}>
-				{routes.map((route) => (
-					<RouteCard imageUrl={null} key={route.id} name={route.name} />
-				))}
-			</ul>
+			<>
+				{locationError?.trim() && (
```
kalinkaaaa14, line 63: Why do we need to use trimmed value?

liza-veis: We need to use booleans in conditions, and I believe there is no need to trim the value, and all errors will already be formatted. But we need to have it like this

```
const hasLocationError = Boolean(locationError);

return hasLocationError && <Component/>
```

If locationError is an empty string, && operator will return an empty string, because it's the first value. And an empty string in jsx is being rendered as an element and can cause unexpected issues on ui

PR #456, opened 2025-08-20
fix(frontend/backend): fix upload max size message and carousel wheel event ss-434

### LINKED ISSUE #434: fix: Carousel does not display newly uploaded images due to 10 items upload limit and carousel line is not visible until hover
### Describe your issue

The image carousel on the Explore page does not update with newly uploaded images. Even though the upload request is successful and files are stored in the system, the carousel continues to display only the first 10 uploaded images. 
Additionally, the carousel line itself is not visible unless the user hovers over the area.

### Steps to reproduce

1. Log in to the website
2. Navigate to Explore page
3. Find the carousel with images (files)
4. Upload more than 10 images
5. Check the carousel display

The carousel only displays the first 10 uploaded images. Newly uploaded files are stored successfully but are not visible in the carousel.
The carousel navigation line is invisible by default and becomes visible only on hover.

The carousel should display the most recently uploaded images, replacing the older ones.
The carousel navigation line should always be visible (not only on hover)

https://binarystudioacademyhq.slack.com/files/U08UE1D1AHJ/F09AXL5HNAZ/select_file.mp4

Windows

Chrome

### Severity

Minor (S4)

vyacheslavKhoroshev:
- fix upload max size message 
- carousel wheel event

Resolves #434 

https://github.com/user-attachments/assets/6ee8aca8-062b-403f-92ee-171890e73382

### DIFF
--- a/apps/backend/src/modules/files/libs/enums/file-exeption-message.enum.ts
+++ b/apps/backend/src/modules/files/libs/enums/file-exeption-message.enum.ts
@@ -1,6 +1,10 @@
+import { config } from "~/libs/modules/config/config.js";
+
+const MAX_FILE_SIZE = String(config.ENV.AWS.MAX_FILE_SIZE_MB);
+
 const FileExceptionMessage = {
 	FILE_REQUIRED: "File is required.",
-	FILE_SIZE_EXCEEDS_LIMIT: "File size exceeds the maximum limitMB.",
+	FILE_SIZE_EXCEEDS_LIMIT: `File size exceeds the maximum limit of ${MAX_FILE_SIZE}Mb.`,
 	INVALID_FILE_TYPE:
 		"Invalid file type. Only JPEG, JPG, and PNG files are allowed.",
 } as const;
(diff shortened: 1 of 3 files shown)


apps/backend/src/modules/files/libs/enums/file-exeption-message.enum.ts
```diff
@@ -1,6 +1,10 @@
+import { config } from "~/libs/modules/config/config.js";
+
+const MAX_FILE_SIZE = String(config.ENV.AWS.MAX_FILE_SIZE_MB);
+
 const FileExceptionMessage = {
 	FILE_REQUIRED: "File is required.",
-	FILE_SIZE_EXCEEDS_LIMIT: "File size exceeds the maximum limitMB.",
+	FILE_SIZE_EXCEEDS_LIMIT: `File size exceeds the maximum limit of ${MAX_FILE_SIZE}Mb.`,
```
lavandosya, line 7: A5 — The names of methods/functions and properties/variables of objects should not contain the names of the object/module.)

liza-veis (approved):

PR #461, opened 2025-08-20
feat(frontend/backend/shared): add an ability to leave a review comment for the route ss-389

### LINKED ISSUE #389: feat: Add an ability to leave a review comment for the route
### What feature?

- [ ] Add reviews section to the routes page and load route reviews.
- [ ] Authorized users should see Add review button
- [ ] When the user clicks Add review button, a modal should be opened, the same as for POI creation, but with the only field - comment.
- [ ] After the creation, the user should receive a success toast notification, and the comment should be added to the top of the list.

### Screenshots

Image

vldkorin:
Added

Backend
- Extended `GET /reviews` to accept optional query parameters

Frontend
- Implemented route reviews listing and review creation

Components added
- ReviewCard - displays reviews as cards;
- CreateReviewModal - modal to create a review;
- RouteReviewsSection - section on the route details page;

Behavior
- Authorized users see an “Add review” button; clicking it opens the modal;
- After a successful creation, users receive a toast notification

Video example

https://github.com/user-attachments/assets/c97d461a-979e-46f9-a51e-269a17cea9e2

Closes #389

### DIFF
--- a/apps/frontend/src/modules/routes/slices/actions.ts
+++ b/apps/frontend/src/modules/routes/slices/actions.ts
@@ -3,12 +3,6 @@ import { createAsyncThunk } from "@reduxjs/toolkit";
 import { StorageKey } from "~/libs/modules/storage/storage.js";
 import { toastNotifier } from "~/libs/modules/toast-notifier/toast-notifier.js";
 import {
-	type ReviewGetAllSearchQuery,
-	type ReviewGetByIdResponseDto,
-	type ReviewRequestDto,
-} from "~/modules/reviews/reviews.js";
-
-import {
 	type APIResponse,
 	type AsyncThunkConfig,
 	type PaginationMeta,
@@ -17,6 +11,11 @@ import {
 	type PointsOfInterestGetAllItemResponseDto,
 	type PointsOfInterestGetAllQuery,
 } from "~/modules/points-of-interest/points-of-interest.js";
+import {
+	type ReviewGetAllSearchQuery,
+	type ReviewGetByIdResponseDto,
+	type ReviewRequestDto,
+} from "~/modules/reviews/reviews.js";
 
 import { RouteNotification } from "../libs/enums/enums.js";
 import {
@@ -145,7 +144,7 @@ const createReview = createAsyncThunk<
 	toastNotifier.showSuccess("Review created successfully.");
 
 	return review;
-
+});
 const constructRoute = createAsyncThunk<
 	APIResponse<PlannedPathResponseDto>,
 	RouteConstructRequestDto,


kalinkaaaa14:
@vldkorin Check please two things here
- other users can see the reviews (not only their own)
- the avatars are loaded successfully on reviews block, the functionality of uploading it is already merged - attach screenshot please

apps/frontend/src/pages/route-description/libs/components/route-reviews/route-reviews.tsx
```diff
@@ -0,0 +1,78 @@
...
+	routeId: number;
+};
+
+const RouteReviews = ({ routeId }: Properties): React.JSX.Element => {
+	const dispatch = useAppDispatch();
+
+	const isAuthenticatedUser = useAppSelector(({ auth }) =>
+		Boolean(auth.authenticatedUser),
+	);
+	const items = useAppSelector((s) => s.review.items);
```
kalinkaaaa14, line 30:
```suggestion
	const items = useAppSelector(({ review }) => review.items);
```

apps/frontend/src/pages/route-description/libs/components/route-reviews/route-reviews.tsx
```diff
@@ -0,0 +1,78 @@
...
+	const [isCreateReviewOpen, setIsCreateReviewOpen] = useState<boolean>(false);
+
+	const openModal = useCallback((): void => {
+		setIsCreateReviewOpen(true);
+	}, []);
+
+	const closeModal = useCallback((): void => {
+		setIsCreateReviewOpen(false);
+	}, []);
```
kalinkaaaa14, line 39:
```suggestion
const handleModalToggle = useCallback(() => {
		setIsCreateReviewOpen((previous) => !previous);
	}, []);
```

apps/frontend/src/pages/route-description/libs/components/route-reviews/route-reviews.tsx
```diff
@@ -0,0 +1,78 @@
...
+	);
+
+	return (
+		<section className={styles["section"]}>
+			<div className={styles["header"]}>
+				<h2 className={styles["title"]}>Reviews</h2>
+				{isAuthenticatedUser && (
+					<div>
+						<Button label="Leave review" onClick={openModal} />
+					</div>
```
kalinkaaaa14, line 60:
```suggestion
						<Button label="Leave review" onClick={openModal} />
```

apps/frontend/src/modules/reviews/slices/review.slice.ts
```diff
@@ -0,0 +1,50 @@
...
+
+type State = {
+	data: null | ReviewGetByIdResponseDto;
+	dataStatus: ValueOf<typeof DataStatus>;
+	items: ReviewGetByIdResponseDto[];
+};
+
+const initialState: State = {
+	data: null,
+	dataStatus: DataStatus.IDLE,
```
kalinkaaaa14, line 17: we should have separate statuses for fetching all reviews and create new one

apps/frontend/src/libs/components/review-card/review-card.tsx
```diff
@@ -0,0 +1,43 @@
...
+			key: "users",
+			name: "Users",
+			permissions: [],
+		},
+		groupId: 0,
+		id: review.user.id,
+		isVisibleProfile: true,
+
+		lastName: review.user.lastName,
+	} satisfies UserAuthResponseDto & { avatarUrl?: null | string };
```
kalinkaaaa14, line 27: Should be removed as we have already user with avatarUrl

apps/frontend/src/libs/components/create-review-modal/create-review-modal.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const CreateReviewModal = ({
+	isOpen,
+	onClose,
+	onSubmit,
+	routeId,
+}: Properties): React.JSX.Element => {
+	const { control, errors, handleReset, handleSubmit } =
+		useAppForm<ReviewRequestDto>({
+			defaultValues: { content: "", poiId: null, routeId },
```
kalinkaaaa14, line 25:
```suggestion
			defaultValues: DEFAULT_CREATE_REVIEW_PAYLOAD
```

apps/backend/src/modules/reviews/review.service.ts
```diff
@@ -27,7 +28,9 @@ class ReviewService implements Service {
 	}
 
 	public async create(
-		payload: ReviewCreatePayload,
+		payload: ReviewCreatePayload & {
+			currentUser: { firstName: string; id: number; lastName: string };
+		},
```
kalinkaaaa14, line 33:
```suggestion
		payload: ReviewCreatePayload, currentUser: UserAuthResponseDto
```

apps/backend/src/modules/reviews/review.repository.ts
```diff
@@ -15,23 +20,42 @@ class ReviewRepository implements Repository {
...
+	public async findAll(
+		options: null | ReviewSearchQuery,
+	): Promise<ReviewGetByIdResponseDto[]> {
+		const q = this.reviewModel
+			.query()
+			.withGraphJoined("user(selectBasic)")
+			.modifiers({
+				selectBasic(b) {
+					b.select("id", "first_name as firstName", "last_name as lastName");
+				},
```
kalinkaaaa14, line 39: It should work like this 🤔 
```suggestion
				selectUserInfo(builder) {
					builder.select("id", "firstName", "lastName");
				},
```

apps/backend/src/modules/reviews/review.repository.ts
```diff
@@ -15,23 +20,42 @@ class ReviewRepository implements Repository {
...
 
 		return ReviewEntity.initialize(review);
 	}
 
-	public async findAll(): Promise<ReviewEntity[]> {
-		const reviews = await this.reviewModel.query().execute();
+	public async findAll(
+		options: null | ReviewSearchQuery,
+	): Promise<ReviewGetByIdResponseDto[]> {
+		const q = this.reviewModel
```
kalinkaaaa14, line 33: try to avoid single letter variables

apps/backend/src/modules/reviews/review.model.ts
```diff
@@ -16,6 +31,8 @@ class ReviewModel extends AbstractModel {
 
 	public routeId!: null | number;
 
+	public user!: { firstName: string; id: number; lastName: string };
+
```
kalinkaaaa14, line 35: user: Pick<UserGetByIdItemResponseDto, "firstName" | "id" | "lastName">;

apps/backend/src/modules/reviews/review.model.ts
```diff
@@ -1,9 +1,24 @@
...
 import {
 	AbstractModel,
 	DatabaseTableName,
 } from "~/libs/modules/database/database.js";
 
+import { UserModel } from "../users/user.model.js";
+
 class ReviewModel extends AbstractModel {
+	public static readonly relationMappings = {
```
kalinkaaaa14, line 11: Let's place relation mapping at the end of the model

apps/frontend/src/libs/components/review-card/styles.module.css
```diff
@@ -0,0 +1,25 @@
+.card {
+	gap: 20px;
+	background: var(--color-base);
+}
+
+.head {
+	display: flex;
+	gap: 20px;
+	align-items: center;
+	min-width: 171px;
```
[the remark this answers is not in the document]
liza-veis: On design, it just hugs the content, so it takes the content size. But I'm not sure if we even need this. I would understand if there was a max-width constraint, to break the content when it overflows, but as for min-width, I don't see a reason for adding it

apps/backend/src/modules/reviews/review.service.ts
liza-veis: We don't even need to pass currentUser here; we only need to pass the user ID here

apps/backend/src/modules/reviews/review.repository.test.ts
```diff
@@ -38,21 +36,129 @@ describe("ReviewRepository", () => {
...
+			_captured?: { column?: string; value?: unknown };
+			execute(): Promise<ReviewGetByIdResponseDto[]>;
+			modifiers(map: Record<string, (b: unknown) => void>): FakeQB;
+			orderBy(col: string, directory: "asc" | "desc"): FakeQB;
+			select(...cols: string[]): FakeQB;
+			where(col: string, value: unknown): FakeQB;
+			withGraphJoined(spec: string): FakeQB;
+		};
+
+		const fakeQb: FakeQB = {
```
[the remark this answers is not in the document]
vldkorin:
```
If you are facing issues with withGraphJoined
```
Yes, i'll take a look

apps/backend/src/modules/reviews/review.service.ts
vldkorin: @liza-veis as I understand that I need to return the response status and model to the previous one, where there was only userId, but how do I extract the avatar, first, and last name?

apps/frontend/src/libs/components/review-card/review-card.tsx
vldkorin: @kalinkaaaa14 actually, we we need to pass full UserAuthResponseDto to avatar?

vldkorin:
@kalinkaaaa14 


https://github.com/user-attachments/assets/9010890b-7e5f-47f2-b930-7b6e0466e2a0

apps/backend/src/modules/reviews/review.repository.ts
```diff
@@ -15,23 +21,64 @@ class ReviewRepository implements Repository {
...
+				},
+			})
+			.select(
+				"reviews.id as id",
+				"reviews.content as content",
+				"reviews.likes_count as likesCount",
+				"reviews.poi_id as poiId",
+				"reviews.route_id as routeId",
+			)
+			.orderBy("reviews.id", "desc");
```
liza-veis, line 58: We already have a enum for this, and it's better to order by created_at (from newest to oldest)
```suggestion
			.orderBy("reviews.created_at.", SortOrder.DESC);
```

apps/backend/src/modules/reviews/review.repository.ts
```diff
@@ -15,23 +21,64 @@ class ReviewRepository implements Repository {
...
+
+		const routeId = options?.routeId;
```
liza-veis, line 60: let's extract it at the top

```ts
const { routeId } = options ?? {};
```

apps/backend/src/modules/reviews/review.service.ts
```diff
@@ -27,7 +32,11 @@ class ReviewService implements Service {
 	}
 
 	public async create(
-		payload: ReviewCreatePayload,
+		payload: Pick<
+			UserAuthResponseDto,
+			"avatarUrl" | "firstName" | "id" | "lastName"
+		> &
+			ReviewCreatePayload,
```
liza-veis, line 39: We need to pass it as a user object, just with limited properties
```suggestion
		payload:  {
		   user: Pick<
			UserAuthResponseDto,
			"avatarUrl" | "firstName" | "id" | "lastName"
		>} &
			ReviewCreatePayload,
```

apps/frontend/src/libs/components/review-card/styles.module.css
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.username {
+	font-size: var(--font-size-small);
+	line-height: 1.5;
+	color: var(--color-text-strong);
+}
+
+.content {
+	gap: 20px;
```
liza-veis, line 20: Let's use either 16 or 24 for gaps. We mostly prefer everything to be divisible by 8, but sometimes we may also need 4 or 12

apps/frontend/src/modules/reviews/slices/review.slice.ts
liza-veis, line 1: We don't need a separate slice for reviews, because it's not a separate page. We need to keep reviews for route details in the route-details slice. In this module, we will only keep the api

apps/frontend/src/pages/route-description/libs/components/route-reviews/route-reviews-section.tsx
```diff
@@ -0,0 +1,72 @@
...
+	routeId: number;
+};
+
+const RouteReviewsSection = ({ routeId }: Properties): React.JSX.Element => {
+	const dispatch = useAppDispatch();
+
+	const isAuthenticatedUser = useAppSelector(({ auth }) =>
+		Boolean(auth.authenticatedUser),
+	);
+	const items = useAppSelector(({ review }) => review.items);
```
liza-veis, line 30: It's just a section, let's pass state and actions through props. This will also make it more reusable for the future. It's better to keep logic on the above level and avoid accessing it directly in smaller components. useEffect can also be moved to the page itself.

packages/shared/src/modules/reviews/libs/types/review-get-all-response-dto.type.ts
```diff
@@ -0,0 +1,15 @@
+type ReviewGetAllResponseDto = {
+	content: string;
+	id: number;
+	likesCount: number;
+	poiId: null | number;
+	routeId: null | number;
+	user: {
+		avatar: null | { url?: null | string };
```
liza-veis, line 8: let's return either string or null
```suggestion
		avatarUrl: null | string;
```

packages/shared/src/modules/reviews/libs/types/review-get-all-response-dto.type.ts
```diff
@@ -0,0 +1,15 @@
+type ReviewGetAllResponseDto = {
```
liza-veis, line 1: Get all response dto is array. To define a type for item in it, we add item suffix
```suggestion
type ReviewGetAllItemResponseDto = {
```

packages/shared/src/modules/reviews/libs/types/review-search-query.type.ts
```diff
@@ -0,0 +1,5 @@
+type ReviewSearchQuery = {
```
liza-veis, line 1: Let's also add an operation here, the search query may differ for different methods
```suggestion
type ReviewGetAllSearchQuery = {
```

apps/backend/src/modules/reviews/review.repository.ts
```diff
@@ -15,23 +21,64 @@ class ReviewRepository implements Repository {
...
+
+		if (typeof routeId === "number") {
+			queryBuilder.where("reviews.route_id", routeId);
+		}
+
+		const rows =
+			(await queryBuilder.execute()) as unknown as ReviewGetAllResponseDto[];
```
liza-veis, line 67: why do we need to cast the result here, we actually can rely on response dto interface in repository level
```suggestion
			(await queryBuilder.execute());
```

packages/shared/src/modules/reviews/libs/types/review-get-all-response-dto.type.ts
vldkorin: It’s a derived property that we build in the service layer by mapping avatar.url into a flat avatarUrl field for the dto

packages/shared/src/modules/reviews/libs/types/review-get-all-item-response-dto.type.ts
```diff
@@ -0,0 +1,15 @@
...
+	id: number;
+	likesCount: number;
+	poiId: null | number;
+	routeId: null | number;
+	user: {
+		avatar: null | { url?: null | string };
+		firstName: string;
+		id: number;
+		lastName: string;
+	};
```
kalinkaaaa14, line 12: Can we describe like this?
```suggestion
		user: Pick<
		UserAuthResponseDto,
		"avatarUrl" | "firstName" | "id" | "lastName"
	>;
```

apps/frontend/src/libs/components/review-card/review-card.tsx
```diff
@@ -0,0 +1,32 @@
...
+	review: ReviewGetByIdResponseDto;
+};
+
+const ReviewCard = ({ review }: Properties): React.JSX.Element => {
+	const avatarUser = {
+		avatarUrl: review.user.avatarUrl,
+		firstName: review.user.firstName,
+		id: review.user.id,
+		lastName: review.user.lastName,
+	};
```
kalinkaaaa14, line 16: You don't need this object - review.user is already in this format

apps/frontend/src/libs/components/review-card/review-card.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+	return (
+		<li className={styles["card"]}>
+			<div className={styles["head"]}>
+				<Avatar size={48} user={avatarUser} />
```
kalinkaaaa14, line 21:
```suggestion
				<Avatar size={48} user={review.user} />
```

apps/backend/src/modules/reviews/review.service.test.ts
```diff
@@ -72,8 +76,16 @@ describe("ReviewService", () => {
 		userId: 10,
 	};
 
+	const currentUser = {
```
kalinkaaaa14, line 79:
```suggestion
	const currentUser: Type = {
```

apps/backend/src/modules/reviews/review.repository.ts
```diff
@@ -15,23 +21,49 @@ class ReviewRepository implements Repository {
...
+			.select(
+				"reviews.id as id",
+				"reviews.content as content",
+				"reviews.likes_count as likesCount",
+				"reviews.poi_id as poiId",
+				"reviews.route_id as routeId",
+			)
+			.orderBy("reviews.createdAt", SortingOrder.DESC);
+
+		if (typeof routeId === "number") {
```
kalinkaaaa14, line 62:
```suggestion
		if (routeId) {
```

apps/backend/src/modules/reviews/review.repository.ts
```diff
@@ -15,23 +21,49 @@ class ReviewRepository implements Repository {
...
+		const queryBuilder = this.reviewModel
+			.query()
+			.withGraphJoined("[user(selectUser).avatar(selectAvatar)]")
+			.modifiers({
+				selectAvatar(builder) {
+					builder.clearSelect().select("files.url as url");
+				},
+				selectUser(builder) {
+					builder
+						.clearSelect()
```
kalinkaaaa14, line 45: just wondering why it's needed?

packages/shared/src/modules/reviews/libs/types/review-get-all-item-response-dto.type.ts
vldkorin: https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/461#discussion_r2292225621

apps/backend/src/modules/reviews/review.repository.test.ts
```diff
@@ -7,7 +7,7 @@ import { ReviewEntity } from "./review.entity.js";
 import { ReviewModel } from "./review.model.js";
 import { ReviewRepository } from "./review.repository.js";
 
-describe("ReviewRepository", () => {
+describe("ReviewRepository (knex-mock-client, no fakeQB)", () => {
```
liza-veis:
```suggestion
describe("ReviewRepository", () => {
```

apps/frontend/src/modules/route-details/route-details-api.ts
```diff
@@ -0,0 +1,60 @@
...
+	type ReviewRequestDto,
+} from "./libs/types/types.js";
+
+type Constructor = {
+	baseUrl: string;
+	http: HTTP;
+	storage: Storage;
+};
+
+class RouteDetailsApi extends BaseHTTPApi {
```
liza-veis, line 22: This is the reviews api, and it should be in the reviews module. We group APIs by feature, while slices represent information about a specific page, because we may load different types of the same type on multiple pages in different formats. So even though we load reviews for the route details slice, the reviews api itself should be in a separate reviews module

apps/frontend/src/pages/route-description/route-details.tsx
```diff
@@ -76,6 +87,32 @@ const RouteDetails = (): React.JSX.Element => {
 		}
 	}, [route, handleValueSet]);
 
+	const [reviews, setReviews] = useState<ReviewGetByIdResponseDto[]>([]);
```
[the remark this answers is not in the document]
liza-veis: Nope, reviews should not be taken from the local state, but from the route details slice, while the reviews api should remain in the reviews module. We can't do any async logic in components and await anything or call the toast notifier

liza-veis (changes requested):

apps/backend/src/modules/reviews/review.repository.ts
```diff
@@ -15,23 +18,64 @@ class ReviewRepository implements Repository {
...
+			.orderBy("reviews.createdAt", SortingOrder.DESC);
+
+		if (routeId) {
+			queryBuilder.where("reviews.route_id", routeId);
+		}
+
+		const rows = await queryBuilder.execute();
 
-		return reviews.map((review) => ReviewEntity.initialize(review));
+		return rows.map((r) =>
```
liza-veis, line 63: let's use more clear naming
```suggestion
		const reviews = await queryBuilder.execute();

		return reviews.map((review) =>
```

liza-veis (approved):
LGTM, one small comment ✨

liza-veis (approved):
Please resolve merge conflicts 🐈

kalinkaaaa14 (approved):

liza-veis (approved):

PR #479, opened 2025-08-21
fix(frontend): update map location logic ss-474

### LINKED ISSUE #474: fix: Cannot choose location on the map after POI modal submit rejection
### Describe your issue

Cannot choose location on the map after POI modal submit rejection

### Steps to reproduce

1. Go to the https://smart-scapes.com/app/manage-routes?poi-page=1&poi-page-size=10 as admin
2. Open POI modal
3. Enter POI name that has already exist and choose the location
4. Submit the form and see the rejection
5. Try to choose one more time location after unsuccessful submission

User cannot select a new POI location

User should be able to interact with the map and select a location again after a failed submission

Image

Windows

Chrome

### Severity

Major (S3)

vldkorin:
Fix #474 

https://github.com/user-attachments/assets/eb943ea5-ff60-401e-bb49-aa1713d44695

Closes #474

### DIFF
--- a/apps/frontend/src/libs/components/map-location-field/map-location-logic/map-location-logic.tsx
+++ b/apps/frontend/src/libs/components/map-location-field/map-location-logic/map-location-logic.tsx
@@ -26,6 +26,12 @@ const MapLocationLogic = ({ location, onLocationChange }: Properties): null => {
 
 	const handleAddMarker = useCallback(
 		(coordinates: PointGeometry["coordinates"]) => {
+			if (markerReference.current) {
+				markerReference.current.setCoordinates(coordinates);
+
+				return;
+			}
+
 			markerReference.current = mapClient.addMarker({
 				coordinates,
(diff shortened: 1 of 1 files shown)


apps/frontend/src/libs/components/map-location-field/map-location-logic/map-location-logic.tsx
```diff
@@ -26,6 +26,12 @@ const MapLocationLogic = ({ location, onLocationChange }: Properties): null => {
 
 	const handleAddMarker = useCallback(
```
liza-veis, line 27: i think in this case it's better to rename it to set
```suggestion
	const handleSetMarker = useCallback(
```

vldkorin: Agree, done
@liza-veis

liza-veis (approved):

PR #482, opened 2025-08-21
feat(frontend/shared): update profile page design ss-448

D1monB:
- Updated tabs and profile page design
- Displayed email as a disabled input, aligned with other fields in the profile form
- Implemented a custom checkbox styled with the brand color
- Updated Avatar:
   When no avatar is present, initials are now white ( only in profile not in header )
   Initials are much bigger for the larger profile avatar size

### DIFF
--- a/apps/frontend/src/libs/components/avatar/styles.module.css
+++ b/apps/frontend/src/libs/components/avatar/styles.module.css
@@ -22,16 +22,7 @@
 	justify-content: center;
 	width: 100%;
 	height: 100%;
+	color: var(--color-base);
 	background-color: var(--color-brand-800);
 	border-radius: 50%;
-}
-
-.dark {
-	font-weight: var(--font-weight-semibold);
-	color: var(--color-text-strong);
-}
-
-.light {
-	font-size: calc(var(--font-size-h1));
-	color: var(--color-base);
 }
--- a/apps/frontend/src/pages/profile/libs/components/user-details/user-details.tsx
+++ b/apps/frontend/src/pages/profile/libs/components/user-details/user-details.tsx
@@ -62,7 +62,7 @@ const UserDetails = (): null | React.JSX.Element => {
 	return (
 		<div className={styles["user-details-container"]}>
 			<div className={styles["user-details-avatar"]}>
-				<Avatar size={128} theme="light" user={authenticatedUser} />
+				<Avatar size={128} user={authenticatedUser} />
 				<div className={styles["user-upload-buttons"]}>
 					<Button
 						label="Upload avatar"
--- a/apps/frontend/src/libs/components/avatar/avatar.tsx
+++ b/apps/frontend/src/libs/components/avatar/avatar.tsx
@@ -1,4 +1,4 @@
-import { combineClassNames, getUserInitials } from "~/libs/helpers/helpers.js";
+import { getUserInitials } from "~/libs/helpers/helpers.js";
 import { type UserAuthResponseDto } from "~/modules/users/users.js";
 
 import styles from "./styles.module.css";
@@ -7,7 +7,6 @@ const DEFAULT_AVATAR_SIZE = 32;
 
 type Properties = {
 	size?: number;
-	theme?: "dark" | "light";
 	user: Pick<UserAuthResponseDto, "avatarUrl" | "firstName" | "lastName">;
 };
 
@@ -15,7 +14,6 @@ const FONT_SIZE_RATIO = 3;
 
 const Avatar = ({
 	size = DEFAULT_AVATAR_SIZE,
-	theme = "dark",
 	user,
 }: Properties): React.JSX.Element => {
 	const { avatarUrl, firstName, lastName } = user;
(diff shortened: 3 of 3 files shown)


apps/frontend/src/pages/profile/libs/components/dashboard-heading/styles.module.css
```diff
@@ -0,0 +1,19 @@
+.heading {
+	margin: 0 0 clamp(24px, 6vw, 43px);
+}
+
+.title {
+	margin: 0 0 8px;
+	font-size: var(--font-size-h1);
+	font-weight: var(--font-weight-semibold);
+	line-height: 1.2;
+	color: #1a3f34;
```
my-derkach, line 10: It is better to define or used already defined variable

apps/frontend/src/pages/profile/libs/components/dashboard-heading/dashboard-heading.tsx
my-derkach, line 1: Is it really necessary to have such component?

apps/frontend/src/pages/profile/libs/components/dashboard-heading/styles.module.css
```diff
@@ -0,0 +1,19 @@
+.heading {
+	margin: 0 0 clamp(24px, 6vw, 43px);
```
my-derkach, line 2: I think it would be better to not use clamp and just make this in responsiveness part, especially when your preferred value in vw

apps/frontend/src/pages/profile/libs/components/dashboard-heading/dashboard-heading.tsx
D1monB: I think no, i just took it from manage-routes task, i did there the same component, 
Just same styles

apps/frontend/src/pages/profile/libs/components/dashboard-heading/styles.module.css
D1monB: Do you think it's worth it? It's just not pixel perfect, it just helps avoid @media overwriting the same field, + I also don't have a big screen if I use only vw other users may have different margins, I just set maximum value "43" to be like in the layout

apps/frontend/src/pages/profile/libs/components/dashboard-heading/dashboard-heading.tsx
my-derkach: Oh, ok, then no problem with that

my-derkach (approved):
Looks good to me

apps/frontend/src/libs/components/checkbox/styles.module.css
```diff
@@ -1,6 +1,24 @@
 .container {
 	display: flex;
-	column-gap: 20px;
+	column-gap: 15px;
```
liza-veis, line 3:
```suggestion
	column-gap: 16px;
```

liza-veis (changes requested):

apps/frontend/src/pages/profile/libs/components/user-details/user-details.tsx
```diff
@@ -62,20 +62,22 @@ const UserDetails = (): null | React.JSX.Element => {
...
+						label="Upload avatar"
+						onClick={handleTriggerFileUpload}
+						variant="outlined"
+					/>
+					<Button label="Delete avatar" variant="outlined-danger" />
+					<input
+						accept="image/*"
+						onChange={handleFileUpload}
+						ref={fileInputReference}
+						style={{ display: "none" }}
```
liza-veis, line 77: let's use clasName here
```suggestion
```

apps/frontend/src/pages/profile/libs/components/dashboard-heading/dashboard-heading.tsx
```diff
@@ -0,0 +1,20 @@
+import styles from "./styles.module.css";
+
+type Properties = {
+	subtitle: string;
+	title: string;
+};
+
+const DashboardHeading = ({
```
liza-veis, line 8: we already have PageHeading common component, let's use it instead

```suggestion
```

apps/frontend/src/libs/components/avatar/styles.module.css
```diff
@@ -21,9 +21,16 @@
...
+
+.small {
+	font-weight: var(--font-weight-semibold);
+	color: var(--color-text-strong);
+}
+
+.large {
+	font-size: calc(var(--font-size-h1));
+	color: var(--color-base);
+}
```
liza-veis, line 36: Small and large are size options are not style variants, so they are in conflict with size property that we have. And it's very confusing and hard to predict the behavior from outside when we have both size properties and variant small/large which is responsible for size. 

1. We either need to get rid of the numeric value for size and use fixed size options - `size: small | large`. Then we will need to update it across the project.
2. Or we need to keep the size, but then we need to calculate font-size dynamically, e.g.
```tsx
  const fontSize = Math.floor(size / 2); // or any ratio that looks good
  const avatarStyle = { "--font-size": `${fontSize}px`} as React.CSSProperties;

  return (
    <div
      className={styles["avatar"]}
      style={avatarStyle}
    >
``` 
and then apply this var to font size in css

liza-veis (changes requested):

liza-veis (approved):

PR #490, opened 2025-08-21
feat(frontend): user routes flow ss-347

### LINKED ISSUE #347: feat: Implement user routes flow on backend
### What feature?

Add an ability for the user to start a route, see it as “active”, and finish it by sending the actual path. Only one active route per user at a time.

- [x] Create table user_routes
```
Table user_routes {
  id integer [primary key]
  created_at timestamp
  updated_at timestamp
  route_id integer [not null]
  user_id integer [not null]
  planned_path geometry(LineString)
  actual_path geometry(LineString)
  started_at timestamp
  completed_at timestamp
  status enum // not_started, active, completed, cancelled, expired [not null]
}
```
- [x] Start the route
   - `POST /user-routes`
   - Body: `{ "routeId": number }`
   - If the user already has an active user_route → 409 Conflict.
   - Load route by routeId, read its planned geometry, snapshot into planned_path.
   - Insert user_routes row with status=active, return created record
  
- [x] Finish the route
  - `POST /user-routes/:id/finish`
  - Body
  
```
      {
        "actualPath": { "type": "LineString", "coordinates": [[lon,lat], ...] }
      }
```

  - User route must be owned by the user and be currently active
     - not found → 404
     - not owned → 403
     - wrong status → 409

### Screenshots

_No response_

vyacheslavKhoroshev:
https://github.com/user-attachments/assets/e73e19eb-057e-45b0-a28c-93c1ba7161d0

Resolves #347


apps/backend/src/modules/user-routes/user-route.service.ts
```diff
@@ -0,0 +1,175 @@
...
+		userRoute: UserRouteResponseDto,
+		userId: number,
+	): void {
+		if (userRoute.userId !== userId) {
+			throw new UserRouteError({
+				message: UserRouteExeptionMessage.USER_ROUTE_NOT_OWNED,
+				status: HTTPCode.FORBIDDEN,
+			});
+		}
+	}
```
liza-veis, line 159: I would rather not have this as a separate check. But if you want to keep it, it's better to avoid functions that accept the whole interface and compare or use it as it wants, while actually only a few properties from the interface are needed. If we want to keep it, instead of passing userRoute, we need to pass userRouteUserId. 

Why keeping it this way may not be the best option in terms of maintenance and extendability?

Our interface for specific operations may change at any time, and while it was accessible by userId here, this field may have a different location for other methods. + We will need to keep track of all interfaces and always consider how changes in them will impact such helpers. Ideally, we don't want to even think about the helper's internal implementation when we update the interface.

packages/shared/src/libs/validated-schemas/coordinates.validation-schema.ts
```diff
@@ -0,0 +1,41 @@
+import { z } from "zod";
+
+import {
+	CoordinatesValidationMessage,
+	CoordinatesValidationRule,
+} from "../enums/enums.js";
+import { parseToFloat } from "../helpers/helpers.js";
+
+const latitudeSchema = z
```
liza-veis, line 9: I'm ok with keeping exception messages in shared, but as for validation schemas, let's define them separately. Sometimes they are required, sometimes they are not, sometimes they are stored as different data types and have different validation rules. We already had 3 separate cases specifically for the location schema, where we needed something different. And generally allowing multiple options in validation schemas (number/string) will negatively affect troubleshooting/fixing, and reduce extendability in the future. It's better to have two totally the same validation schemas for different parts of the application, than one with conditions and multiple options

apps/backend/src/db/migrations/20250817134209_add_user_routes_table.ts
```diff
@@ -0,0 +1,100 @@
...
+			.inTable(ROUTES_TABLE)
+			.onDelete("CASCADE");
+		table
+			.enum(ColumnName.STATUS, USER_ROUTE_STATUS, {
+				enumName: "user_route_status",
+				useNative: true,
+			})
+			.notNullable();
+		table.timestamp(ColumnName.STARTED_AT).nullable().defaultTo(null);
+		table.timestamp(ColumnName.COMPLETED_AT).nullable().defaultTo(null);
```
kalinkaaaa14, line 67:
```suggestion
		table.timestamp(ColumnName.STARTED_AT).nullable();
		table.timestamp(ColumnName.COMPLETED_AT).nullable();
```

packages/shared/src/modules/user-routes/libs/enums/user-route-validation-message.ts
kalinkaaaa14, line 1: user-route-validation-message.enum.ts

packages/shared/src/libs/enums/coordinates-validation-rule.ts
kalinkaaaa14, line 1: coordinates-validation-rule.enum.ts

apps/backend/src/modules/user-routes/user-route.service.ts
```diff
@@ -0,0 +1,175 @@
...
+
+		const createdData = UserRouteEntity.initializeNew({
+			actualGeometry: geometry,
+			plannedGeometry: geometry,
+			routeId,
+			status: UserRouteStatus.NOT_STARTED,
+			userId,
+		});
+
+		await this.ensureIsNotDublicateRoute(userId, geometry);
```
kalinkaaaa14, line 43:
```suggestion
		await this.ensureIsNotDuplicateRoute(userId, geometry);
```

apps/backend/src/modules/user-routes/user-route.repository.ts
```diff
@@ -0,0 +1,151 @@
...
+					"ST_AsGeoJSON(planned_geometry)::json as planned_geometry",
+				),
+			])
+			.execute();
+
+		return result.map((item) => UserRouteEntity.initialize(item));
+	}
+
+	public async findByRouteId(routeId: number): Promise<null | UserRouteEntity> {
+		const result = await this.userRouteModel
```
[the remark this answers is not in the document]
kalinkaaaa14: or you can use .first()

apps/backend/src/modules/user-routes/user-route.repository.ts
```diff
@@ -0,0 +1,151 @@
...
+			.execute();
+
+		return result[0] ? UserRouteEntity.initialize(result[0]) : null;
+	}
+
+	public async hasActiveRoute(userId: number): Promise<boolean> {
+		const result = await this.userRouteModel
+			.query()
+			.where({ status: UserRouteStatus.ACTIVE, userId })
+			.execute();
```
kalinkaaaa14, line 92:
```suggestion
		const result = await this.userRouteModel
			.query()
			.select(1)
			.where({ status: UserRouteStatus.ACTIVE, userId })
			.execute();
```

packages/shared/src/libs/validated-schemas/coordinates.validation-schema.ts
vyacheslavKhoroshev: @liza-veis

yes, agree, i see your point. But its coordinates always have the same format, it's like a constant. 
About optional, we are using like `latitude: latitudeSchema.optional()` . 
About type, ok, we can create `latitudeStringSchema` and `latitudeNumberSchema`, but i believe we will use them like numbers. 

It just my opinion, so what we will do?)))

apps/backend/src/modules/user-routes/user-route.service.ts
```diff
@@ -0,0 +1,190 @@
...
+			actualGeometry,
+			completedAt: getCurrentIsoDate(),
+			status: UserRouteStatus.COMPLETED,
+		});
+
+		const updatedRoute = (await this.userRouteRepository.patch(
+			userRoute.id,
+			userId,
+			updatedData,
+		)) as UserRouteEntity;
```
liza-veis, line 78: It's better not to cast here, because during patching, we may pass filters that may have been overlooked, which will eventually return null. So it's better to handle this case here and throw a not found error
```suggestion
		));
```

apps/backend/src/modules/user-routes/user-route.repository.ts
```diff
@@ -0,0 +1,127 @@
...
+					"ST_AsGeoJSON(planned_geometry)::json as planned_geometry",
+				),
+			])
+			.execute();
+
+		if (options?.multiple) {
+			return userRoutes.map((item) => UserRouteEntity.initialize(item));
+		}
+
+		return userRoutes[0] ? UserRouteEntity.initialize(userRoutes[0]) : null;
```
liza-veis, line 90: It's better to return a single format, no matter the options. Just return the array from find. In places where you need the first value use `const [userRoute] = userRoutes;`

packages/shared/src/libs/validated-schemas/coordinates.validation-schema.ts
liza-veis: It's not a constant, because we already accept multiple formats here and act depending on them (either string or number), which usually results badly for validation schemas. If we wanted to have them in shared, at least we would've needed to eliminate options so it would be a separate schema for number values and a separate one for strings. Then, as for the coordinateSchema itself, if in some place we would need to make this specific array required or optional, we would need to define it separately. Also, we may have a case where we accept strings and keep them as strings. So I think if the validation schema requires at least one type of different response format (e.g. string/number), it already has to be either split by format type or defined separately for each case. In our case, we can either define two schemas, 1 usual and 1 for query params, or define it separately in each place

liza-veis:
And this comment is still unresolved 🐱  https://github.com/BinaryStudioAcademy/bsa-2025-smartscapes/pull/490/files#r2291866286

PR #522, opened 2025-08-24
feat(frontend/backend): create and display follow notification ss-438

nazartsya:

- created a notification for the user that is being followed with the message "{First name, Last name} started following you."
- displayed bell icon in the header and loaded scrollable notification list (for authorized users)
- displayed relative time
- when the user clicks on notification redirected to /app/users/:id

Screenshot 2025-08-24 112744
Screenshot 2025-08-24 112816


DenisDemirov:
Looks good. But is it ok to load notification by page reload ? It will better to use something like sockets. But maybe its enough for basic version for now.

apps/backend/src/modules/user-follows/user-follows.service.ts
```diff
@@ -30,6 +53,28 @@ class UserFollowsService {
...
+			),
+			entityId: followerId,
+			entityType: NotificationEntityType.USERS,
+			notificationType: NotificationType.USER_FOLLOWED,
+			userId: followingId,
+		};
+
+		await this.notificationService.create(notification);
+
 		await this.userFollowsRepository.followUser(followerId, followingId);
```
liza-veis, line 78: let's switch these two calls in places, just to make sure user is followed before creating notifications

apps/frontend/src/modules/notification/notification-api.ts
```diff
@@ -0,0 +1,36 @@
...
+	http: HTTP;
+	storage: Storage;
+};
+
+class NotificationApi extends BaseHTTPApi {
+	public constructor({ baseUrl, http, storage }: Constructor) {
+		super({ baseUrl, http, path: APIPath.NOTIFICATIONS, storage });
+	}
+
+	public async getNotifications(): Promise<
```
liza-veis, line 21: By quality criteria, we need to avoid entity name repetition for methods if it's clear from the className (NotificationsApi)
```suggestion
	public async getAll(): Promise<
```

liza-veis (changes requested):
Looks good overall, a few small comments 🐈

apps/frontend/src/libs/components/header/libs/components/authenticated-header/libs/components/notifications-popover/notifications-popover.tsx
```diff
@@ -0,0 +1,77 @@
...
+	useState,
+} from "~/libs/hooks/hooks.js";
+import { actions as notificationActions } from "~/modules/notification/notification.js";
+
+import { NotificationItem } from "./libs/components/components.js";
+import styles from "./styles.module.css";
+
+const NotificationsPopover = (): React.JSX.Element => {
+	const [isOpen, setIsOpen] = useState<boolean>(false);
+	const containerReference = useRef<HTMLDivElement>(null);
```
kalinkaaaa14, line 17:
```suggestion
	const containerReference = useRef<HTMLDivElement | null>(null);
```

apps/frontend/src/libs/components/header/libs/components/authenticated-header/libs/components/notifications-popover/notifications-popover.tsx
```diff
@@ -0,0 +1,77 @@
...
+		const handleClickOutside = (event: MouseEvent): void => {
+			if (
+				containerReference.current &&
+				!containerReference.current.contains(event.target as Node)
+			) {
+				setIsOpen(false);
+			}
+		};
+
+		document.addEventListener("mousedown", handleClickOutside);
```
kalinkaaaa14, line 37: We have EventListenerType - add new values here

apps/backend/src/modules/user-follows/libs/enums/follow-notification-message.ts
lavandosya, line 1: follow-notification-message.enum.ts

PR #552, opened 2025-08-25
feat(backend/frontend/shared): implement user bookmarks feature ss-497

h4vefundieyoung:
Updates: 

- Implements a feature that allows a user to save routes to bookmars.
- Implements a feature that allows a user to delete saved routes from bookmarks.

https://github.com/user-attachments/assets/b6bcab12-4f35-4ac8-af95-980f4ce45d66


packages/shared/src/modules/routes/libs/types/route-get-by-id-response-dto.type.ts
```diff
@@ -15,6 +16,10 @@ type RouteGetByIdResponseDto = {
 		name: string;
 		visitOrder: number;
 	}[];
+	userRoute: null | {
+		id: number;
+		status: UserRouteStatusType;
+	};
```
kalinkaaaa14, line 22:
```suggestion
	userRoute: null | Pick<UserRouteResponseDto, 'id' | 'status'>;
```

apps/backend/src/modules/routes/route.controller.test.ts
```diff
@@ -286,7 +288,7 @@ describe("Route controller", () => {
 
 		const updateRouteMockData = {
 			body: {
-				description: updatedRoute.description,
+				description: updatedRoute.description ?? "",
```
kalinkaaaa14, line 291: why is it needed?

apps/frontend/src/modules/routes/slices/routes.slice.ts
```diff
@@ -20,6 +20,7 @@ type State = {
 	dataStatus: ValueOf<typeof DataStatus>;
 	restoreCreateRouteFormStatus: ValueOf<typeof DataStatus>;
 	routes: RouteGetByIdResponseDto[];
+	saveRouteStatus: ValueOf<typeof DataStatus>;
```
kalinkaaaa14, line 23: seems this is not used here, as the value is not changed in this slice

apps/frontend/src/modules/routes/routes-api.ts
```diff
@@ -95,7 +95,7 @@ class RoutesApi extends BaseHTTPApi {
 		const response = await this.load<APIResponse<RouteGetByIdResponseDto>>(
 			this.getFullEndpoint(`${RoutesApiPath.ROOT}${id.toString()}`, {}),
 			{
-				hasAuth: false,
+				hasAuth: true,
```
kalinkaaaa14, line 98: Viewing route page should be available for unauthorized users I guess

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -1,18 +1,27 @@
+import { Icon } from "~/libs/components/components.js";
 import { combineClassNames } from "~/libs/helpers/helpers.js";
+import { useMemo } from "~/libs/hooks/hooks.js";
+import { type IconName } from "~/libs/types/types.js";
 
 import styles from "./styles.module.css";
 
 type Properties = {
+	disabled?: boolean;
```
kalinkaaaa14, line 9:
```suggestion
	isDisabled?: boolean;
```

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -1,18 +1,27 @@
...
+	icon?: IconName;
 	label: string;
 	onClick?: () => void;
+	pressed?: boolean;
```
kalinkaaaa14, line 13: isPressed

apps/backend/src/modules/user-routes/user-route.service.ts
```diff
@@ -156,14 +182,21 @@ class UserRouteService implements Service {
...
+				await this.userRouteRepository.checkHasActiveRoute(userId);
+
+			if (hasActiveRoute) {
+				throw new UserRouteError({
+					message: UserRouteExeptionMessage.USER_ALREADY_ON_ACTIVE_STATUS,
+					status: HTTPCode.CONFLICT,
+				});
+			}
 		}
 	}
```
kalinkaaaa14, line 201: seems the conflicts with main branch were unresolved incorrect, because the correct version should be different, check the main branch please

apps/frontend/src/modules/routes/routes-api.ts
h4vefundieyoung: @kalinkaaaa14 it is, token is optional

apps/backend/src/modules/routes/route.controller.test.ts
h4vefundieyoung: @kalinkaaaa14 we have description as a optional value I believe

apps/backend/src/libs/modules/plugins/auth/auth.plugin.ts
liza-veis, line 1: what do we need this change for 🤔 we don't send user token for white routs in the app, so i'm not sure if we need this

packages/shared/src/modules/user-routes/validation-schemas/user-route-patch.validation-schema.ts
liza-veis, line 1: Is it from your PR and do we need this?

apps/backend/src/modules/user-routes/user-route.controller.ts
```diff
@@ -203,6 +214,67 @@ class UserRouteController extends BaseController {
...
+	 *                   properties:
+	 *                     message:
+	 *                       type: string
+	 *                       example: Unauthorized access
+	 */
+
+	public async delete({
+		params,
+		user,
+	}: APIHandlerOptions<{ params: UserRouteParameters }>): Promise<
```
liza-veis, line 262:
```suggestion
	}: APIHandlerOptions<{ params: UserRouteDeleteParameters }>): Promise<
```

packages/shared/src/modules/user-routes/validation-schemas/user-route-parameters.validation-schema.ts
```diff
@@ -0,0 +1,17 @@
+import { z } from "zod";
+
+import { parseToFloat } from "../../../libs/helpers/helpers.js";
+import { UserRouteValidationMessage } from "../libs/enums/enums.js";
+
+const userRouteParameters = z.object({
```
liza-veis, line 6: considering we have delete, what is this for

apps/backend/src/modules/routes/route.entity.ts
```diff
@@ -58,12 +65,13 @@ class RouteEntity implements Entity {
 		this.description = description;
 		this.pois = pois;
 		this.createdByUserId = createdByUserId;
+		this.userRoute = userRoute?.[0] ?? null;
```
liza-veis, line 68: let's name it as saved user route to make it clearer why it's a single item. 
```suggestion
		this.savedUserRoute = userRoute?.[0] ?? null;
```

apps/backend/src/modules/routes/route.repository.ts
```diff
@@ -138,6 +141,21 @@ class RouteRepository implements Repository {
...
 			})
+			.modify((builder) => {
+				if (userId) {
+					builder.withGraphFetched("userRoute(filtered)").modifiers({
+						filtered(builder) {
+							builder
+								.select(
+									"user_routes.id",
+									"user_routes.status",
+									"user_routes.user_id",
```
liza-veis, line 152: we need to include only not_started routes

apps/backend/src/modules/user-routes/libs/enums/user-route-exception-message.enum.ts
```diff
@@ -1,4 +1,5 @@
 const UserRouteExeptionMessage = {
+	INVALID_ID: "Invalid route id",
```
liza-veis, line 2:
```suggestion
	INVALID_ID: "Invalid route id.",
```

apps/backend/src/modules/user-routes/user-route.repository.ts
```diff
@@ -47,6 +47,39 @@ class UserRouteRepository implements Repository {
 		return UserRouteEntity.initialize(userRoute);
 	}
 
+	public async deleteSavedRoute(id: number, userId: number): Promise<boolean> {
```
liza-veis, line 50: seems like usual delete to me. We only consider not_started routes as saved. Because those that were started will be shown in history and will automatically dissapear from saved 
```suggestion
	public async delete(id: number, userId: number): Promise<boolean> {
```

packages/shared/src/modules/user-routes/validation-schemas/user-route-parameters.validation-schema.ts
h4vefundieyoung: @liza-veis I didn't define it. Most likely merged from the main.

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -1,8 +1,14 @@
...
 import { combineClassNames } from "~/libs/helpers/helpers.js";
+import { useMemo } from "~/libs/hooks/hooks.js";
+import { type IconName } from "~/libs/types/types.js";
 
 import styles from "./styles.module.css";
 
 type Properties = {
+	icon?: IconName;
+	isDisabled?: boolean;
+	isPressed?: boolean;
```
liza-veis, line 11: it's not pressed, it should be a different variant of button, maybe something like ghost
```suggestion
	icon?: IconName;
	isDisabled?: boolean;
```

apps/frontend/src/libs/components/icon/libs/maps/icon-name-to-svg.map.ts
```diff
@@ -27,6 +28,7 @@ const iconNameToSvg: Record<
 	analytics: Analytics,
 	arrowDown: ArrowDown,
 	bell: Bell,
+	bookmarks: Bookmarks,
```
liza-veis, line 31:
```suggestion
	bookmark: Bookmark,
```

packages/shared/src/libs/validated-schemas/coordinates.validation-schema.ts
```diff
@@ -0,0 +1,41 @@
+import { z } from "zod";
+
+import {
+	CoordinatesValidationMessage,
+	CoordinatesValidationRule,
+} from "../enums/enums.js";
+import { parseToFloat } from "../helpers/helpers.js";
+
+const latitudeSchema = z
```
liza-veis, line 9: this file needs to be removed, we get rid of it

packages/shared/src/modules/points-of-interest/libs/validation-schemas/points-of-interest-update.validation-schema.ts
```diff
@@ -15,6 +17,12 @@ const pointOfInterestUpdate = z
 				message: PointsOfInterestValidationMessage.DESCRIPTION_MAXIMUM_LENGTH,
 			})
 			.nullable(),
+		location: z.object({
+			coordinates: coordinateSchema,
+			type: z.literal(LocationType.POINT, {
+				message: PointsOfInterestValidationMessage.INVALID_LOCATION_TYPE,
+			}),
+		}),
```
liza-veis, line 25:
```suggestion
```

apps/backend/src/modules/routes/libs/types/user-route.ts
```diff
@@ -0,0 +1,8 @@
+import { type UserRouteStatusType } from "@smartscapes/shared";
+
+type UserRoute = {
```
liza-veis, line 3: we don't have separate entity types, because we have entities as classes, so we use them for entityType. If this type is only used in entity itself, let's keep in the entity file not to expose it to the module or outside, because entity changes too dynamically. 

BTW I remembered your question about why we are using entities and i forgot to reply to you. We use those entities as our custom models, to use it as connection between service payload and db. So if we want to insert model in db, we use entity and when we retrieve something from repository we return entity. I also don't like how they currently look like, just a common pattern that we've been using for the academy projects. However it's better than using models on service level and they provide a controllable interface

liza-veis (changes requested):

apps/backend/src/libs/modules/plugins/auth/auth.plugin.ts
h4vefundieyoung: @liza-veis I've changed the logic a little bit, so that if token comes user authorizes on the backend whether a route is white listed or not. If a user is authorized - he will get additional data attached to the getById request, that describes his status for that route. The main idea is to prevent an additional request to the server.

apps/backend/src/modules/routes/libs/types/user-route.ts
h4vefundieyoung: @liza-veis i do understand the purpose of entity classes. i've asked about lots of these methods inside entities. it looks like everyone defines methods like a toTwoFieldsFromEntity, toFourPropsFromEntity, toMyLittlePonyFromEntity and whatever - looks messy to me, and i've asked mainly if it should be like this, coz as i understand there should be something like toDataBaseShape() toApplicatonShape(), and all the shape transformanions should be performed in the layers that use these methods, mainly services, as i understand.

apps/backend/src/libs/modules/plugins/auth/auth.plugin.ts
```diff
@@ -49,9 +45,19 @@ const auth = (app: FastifyInstance, { whiteRoutes }: PluginOptions): void => {
...
+			try {
+				await parseToken();
+
+				return;
+			} catch {
+				return;
+			}
 		}
+
+		await parseToken();
```
liza-veis, line 60: seems like it can be simplified to this, also it's better to rename parseToken to something like authorize user, because this function is not only responsible for parsing token, but also finding and attaching user to request
```suggestion
			try {
				await authorizeUser();
		} catch (error) {
		
		       if (!isWhiteListRoute) {
		            trhow error;
			   }
			}
```

apps/backend/src/modules/user-routes/user-route.repository.ts
h4vefundieyoung: @liza-veis the case is signature. delete expects one id argument

apps/backend/src/modules/routes/libs/types/user-route.ts
h4vefundieyoung: @liza-veis gotcha, thanks

liza-veis (approved):

PR #557, opened 2025-08-26
feat(frontend): integrate route creation with construct route flow ss-542

### LINKED ISSUE #542: feat: Integrate route creation with construct route flow
### What feature?

- [x] Add a save button in place of the current construct button (rename to build path) to the construct route page.
- [x]  Save should be disabled unless the planned path is present.
- [x] When clicking save, the user should be redirected to the returnTo route from query params with plannedPathId added to query params
- [x] When clicking on the link icon in the POI card, open the link in a new tab instead of the same tab, to preserve the constructed route

The expected result of the ticket is fully working route creation functionality:
- Create route modal
- Go to the construct page
- Construct route
- Return back to create a modal
- Route creation

Note: The functionality with adding returnTo to the path and taking plannedRouteId from query params is already implemented.

### Screenshots

Image

vldkorin:
Updates:
- **Button component**
  - added support for the isDisabled property to allow disabling buttons when necessary

- **Construct Route page**
  - updated logic to pass isSaveDisabled and plannedRouteId props to the side panel to control the save functionality properly

- **POI Card**
  - modified the link action so that clicking on the POI link now opens it in a new browser tab instead of the same tab
- **Side Panel**
  - added **Save** button.  
  - updated the save handler to persist constructed route data before redirecting to the modal for route creation

https://github.com/user-attachments/assets/ba21e8df-2078-45e0-9e45-aaadd94838c5

Closes #542


apps/frontend/src/pages/construct-route/libs/components/side-panel/side-panel.tsx
```diff
@@ -73,12 +82,31 @@ const SidePanel = ({
...
+
+		void dispatch(
+			preserveCreateRouteFormData({ plannedPathId: plannedRouteId, poiIds }),
+		);
+
+		const [path, query = ""] = route.split("?");
+		const newSearchParameters = new URLSearchParams(query);
+		newSearchParameters.set("plannedRouteId", String(plannedRouteId));
+
+		navigate(`${String(path)}?${newSearchParameters.toString()}`);
```
liza-veis, line 107: we have a common helper, something like getFullUrlWithQuery, let's use it

PR #593, opened 2025-08-28
feat(frontend/backend/shared): display activities and reviews on user public profile ss-545

DenisDemirov:
Screenshot 2025-08-28 162523

https://github.com/user-attachments/assets/a0c59e6c-e014-49fd-8c10-5a8ccf0a8ff8


packages/shared/src/modules/user-routes/validation-schemas/user-route-get-all.validation-schema.ts
```diff
@@ -0,0 +1,16 @@
...
+const userRouteGetAll = z.object({
+	routeId: z
+		.string()
+		.transform(parseToFloat)
+		.pipe(
+			z.number({
+				message: UserRouteValidationMessage.USER_ID_INVALID_TYPE,
+			}),
+		),
+});
```
vldkorin, line 15:
```suggestion
import { parseToInteger } from "../../../libs/helpers/helpers.js";
import { UserRouteValidationMessage } from "../libs/enums/enums.js";

const userRouteGetAll = z.object({
	routeId: z
		.string()
		.transform(parseToInteger)
		.pipe(
			z.number({
				message: UserRouteValidationMessage.USER_ID_INVALID_TYPE,
			}),
		),
});
```

vldkorin (changes requested):

liza-veis (approved):
LGTM

PR #614, opened 2025-08-29
fix(frontend): fix bug that page limit doesnt work for less than default vaule ss-576

DenisDemirov:


timosh-yura:
lgtm

apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -56,11 +56,13 @@ const Table = <T,>({
 			</div>
 
 			<div className={styles["footer"]}>
-				<Pagination
-					paginationSettings={paginationSettings}
-					totalItems={totalItems}
-					totalPages={totalPages}
-				/>
+				{!isLoading && totalItems > 0 && (
```
liza-veis, line 59: not sure about hiding pagination on loading, it will jump every time after switching pages, i would rather keep it

PR #640, opened 2025-08-29
fix(frontend/backend/shared): resolve query string filtering issues ss-620

h4vefundieyoung:
Resolves query string forming issue for public profile/saved-routes/history pages


apps/backend/src/modules/user-routes/user-route.repository.ts
```diff
@@ -84,7 +84,6 @@ class UserRouteRepository implements Repository {
 			.modify((builder) => {
 				if (filters.userId) {
 					builder.where("user_routes.user_id", filters.userId);
-					builder.where("reviews.user_id", filters.userId);
```
liza-veis, line 87: I believe this line should be returned back

